PR #147, opened 2013-11-15
Proposition : Expose custom state of the task in the API

NotSqrt:
In my projects, I use custom states:
http://docs.celeryproject.org/en/latest/userguide/tasks.html#custom-states

A `metadata` dict can be provided, and is stored in AsyncResult.result, just like actual results.

This commit exposes the custom state as a `meta` key.

To filter custom states from the actual result, there does not seem to be another way than checking that the result is not `ready()`. The state could be anything, since it can be changed by the user.

I tried to do the same for the `Task` view, but it appears that the `result` is not kept updated....

Thanks for the awesome work !

### DIFF
--- a/flower/api/tasks.py
+++ b/flower/api/tasks.py
@@ -71,6 +71,8 @@ class TaskResult(BaseTaskHandler):
                                  'traceback': result.traceback})
             else:
                 response.update({'result': self.safe_result(result.result)})
+        else:
+            response.update({'meta': result.result})
         self.write(response)
 
     def safe_result(self, result):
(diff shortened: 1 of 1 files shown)


jheld:
I'm not sure that I see the issue of allowing addition of returning the metadata (I'd expect we would probably wish to use the same key name, though). We can put a warning in the docs discussing the implications of programmatic usage/reliance of the data, but why forbid (again not sure I fully understand the concern from a feature standpoint)?

If we don't intend to merge this, perhaps we should close it?

mher:
Closing this pull request because it is not common use case

PR #199, opened 2014-03-18
Change directory layout of static Javascript files.

bdrung:
Create one directory per project (bootstrap, d3, jquery, jquery-ui) and move all related file in this directory. This makes it easier to track the origin of a static file and to update a library by just replacing one directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
     install_requires=['celery', 'tornado'],
     test_suite="tests",
     tests_require=['mock'],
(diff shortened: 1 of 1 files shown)


pabs3:
Not having the libraries in separate directories makes it a lot harder to update the embedded JavaScript/CSS (see #1078). There is no automatic process for updating them, nor indicator of where the files come from, nor documentation about how to update them.

The amount of libraries has not been going up over time, so the number of subdirectories doesn't seem like it will be high.

PR #209, opened 2014-04-02
adding --basic_auth_file to specify a list of username:login pairs

StevenVanAcker:
Adding --basic_auth_file to specify a list of username:login pairs that (in addition to the user specified with --basic_auth) are allowed access.
This approach is safer when flower is used in a multiuser environment where other users can see the (--basic_auth) credentials in the process list.


ktnr:
Why was this not merged? A basic_auth_file option would be great.

Issue #194, opened 2014-03-13
UX suggestions


arnists:
I hope to have time to devote to this project in the summer, I really like what it offers as a management tool for Celery.  These features would in my opinion improve it's usefulness tremendously.

# /tasks
* while columns are static, should include timestamps (quick)
* columns should be decided by a setting (better)
* columns should be sortable
* time filtering
* bulk revoke from list-view

# /monitor
* async data from pageview (history)
* task filtering

mozillazg:
@sandeepgk1989 
2) flower use CELERY_TIMEZONE to get timezone.

daniel-cohen:
+1 on sorting

meteozond:
+1 for pagination

mher:
Most of the features requested here are implemented in the master version. Please try it.

PR #316, opened 2015-01-15
sort tasks by name/state/received/started

mozillazg:
ref #194


mher:
Merged, thanks! Please add your name to https://github.com/mher/flower/blob/master/AUTHORS

PR #325, opened 2015-02-03
Consistent badges

mgedmin:
Make all three badges look consistent (using the "flat" badge style used by Travis and, since Feb 1, shields.io).

Replace broken crate.io link with a working PyPI link.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,12 +1,12 @@
 Celery Flower
 =============
 
-.. image:: https://badge.fury.io/py/flower.png
-        :target: http://badge.fury.io/py/flower
-.. image:: https://travis-ci.org/mher/flower.png?branch=master
+.. image:: https://pypip.in/version/flower/badge.svg?style=flat
+        :target: https://pypi.python.org/pypi/flower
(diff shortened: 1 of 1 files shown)


mher:
Merged, thanks! Can you add your name to https://github.com/mher/flower/blob/master/AUTHORS?

mgedmin:
I will, as soon as I have something more substantial to contribute :)

mher:
@mgedmin Looks like this commit broke the build status badge. It always shows error. Can you fix it?

mgedmin:
The pypy build broke with a download error while installing dependencies.  As the repo owner you should be able to retry it by clicking the little round circled arrow button on the top-right in https://travis-ci.org/mher/flower/jobs/49546912; as a bystander I cannot do that.

You may also want to change the 'pip install ...' line in .travis.yml to 'travis_retry pip install ...'; that'll make Travis retry failed downloads up to 3 times.  It would also be an opportunity to drop the deprecated --use-mirrors pip flag.

mher:
thanks @mgedmin

PR #342, opened 2015-03-11
Fixing retry interval reset on connection failure

TimOrme:
Retry intervals were not getting reset on connection success. As such, intermittent failures in connectivity would each take longer to retry even if there was a successful connection made between them.


mher:
Thanks for the pull request! Please add your name to AUTHORS.

jackyfkc:
It would be better to limit the maximum number of ***try_interval***

Issue #344, opened 2015-03-12
Show worktime for succeeded tasks in tasks list


Melevir:
For each succeeded task I can calculate work time manually, by subtracting "Started" from "Succeeded", but I have to visit task page for that.
It looks like a good idea to add column with worktime at `/tasks` directly.
This involves API and templates modification.

I'll be glad to try to complete this task once it's approved.

mher:
The /task/ page already contains Runtime column.

If you want to have it on /tasks page probably it makes sense to make it optional. Different users have different use cases. The columns can be added/ removed from UI or from configuration file.

PR #345, opened 2015-03-20
#344 Made tasks columns configurable

Melevir:


mher:
Thanks for the pull request. I think is is better to have all partial temples in a single tasks.html file filtered by `if 'runtime in task_columns` statements.

Melevir:
Ok, I'll fix that.

Melevir:
Done.

mher:
Merged, thanks!

PR #350, opened 2015-03-24
Fix template bug querying rusage when using with celery 3.0

jaysoffian:


mher:
Merged, thanks! Please add your name to the authors.

PR #354, opened 2015-03-26
Add the '-A proj' argument to every flower invokation

brouberol:
I've seen people being confused by the fact that this argument was
sometimes mentioned in the docs, and sometimes now, when it's pretty
central to the flower configuration.


mher:
Merged, thanks!

PR #359, opened 2015-03-30
Fix bug in handling denied emails with OAuth

dellsystem:
With Google OAuth, when a user tries to log in with an email that isn't allowed, a KeyError is triggered in line 46 of flower/views/auth.py because the `user` dictionary does not have an `email` and so the call to `format()` doesn't work. Changing the kwargs to `email=email` fixes that.

I also replaced the HTTPError exception with a call to `self.render()`, because the former does not appear to display any information to the user (causing the request to simply hang) while the latter does. I'm not sure if this is what 404.html is intended for, but it works perfectly for me.

Thanks for your work on flower!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -49,3 +49,4 @@ David Matson
 Paulo SantAnna
 Sanchit Arora
 Ilya Lebedev
+Wendy Liu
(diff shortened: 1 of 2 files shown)


mher:
Thanks for the pull request. Can you only fix the email part? 404.html page is not intended for auth errors.

mher:
Also, please add your name to contributors list

dellsystem:
Sure, I can do that. Is raising HTTPError the intended behaviour, though? I don't know if it's specific to my deployment setup or what, but raising the exception doesn't result in anything being shown to the end user, just an error message to stderr.

mher:
Yes, tornado should catch HTTPError and generate error.html page. If you think it is a bug you can open an issue or another pull request.

dellsystem:
I see. I'm still not able to get it to work, but I'll look into it more to see if I can find a fix. I've changed it back to the HTTPError for now.

mher:
Merged, thanks!

PR #361, opened 2015-03-30
Enable WebSocket CORS

getupcloud:
This commit allows flower to run websocket server listening on a different port than http does, from a client point of view.

This scenario is found when running flower from inside [OpenShift](github.com/openshift/origin-server), where websocket conections are expected on port 8000 instead 80. This causes `tornado.websocket.WebSocketHandler` to fail because method `check_origin`'s default behavior is to allow connections only for matching `Host` and `Origin` HTTP headers.

One can now explicitly specify what are the accepted hosts for websocket connections through option `ws_allowed_hosts`:

```
$ flower --ws_allowed_hosts=http://flower-caruccio.getup.io
```

### DIFF
--- a/flower/options.py
+++ b/flower/options.py
@@ -55,6 +55,8 @@ define("natural_time", type=bool, default=True,
        help="show time in relative format")
 define("tasks_columns", type=str, default="name,uuid,state,args,kwargs,result,received,started",
        help="Slugs of columns on /tasks/ page, delimited by comma")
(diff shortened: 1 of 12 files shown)


getupcloud:
This is a fix PR 'cause previous wasn't properly tested before submission. Sorry for the noise.

mmange:
Can we get this reviewed and merged? currently flower web sockets are not usable

mher:
Please try the master version. WebSockets were removed entirely.

Issue #358, opened 2015-03-30
Search tasks by args/kwargs


Melevir:
Consider a use case: I have 1 million succeeded/failed tasks with "client_id" argument. I want to see list of all tasks, that have "client_id=1234".

For me it would be enough an "args/kwargs" search field on the `/tasks/` page, that I can fill with "1234" and get list of tasks, that have "1234" as value for any argument. Optionally, I want to fill the field with "client_id=1234" value and get tickets with client_id I need.

Is the idea looks nice to you? I can try to implement it if, it is approved.

mher:
Ability to search/filter tasks by args/kwars/results will be very useful. Thanks!

mher:
This feature is supported now. Try the master version.

PR #364, opened 2015-04-01
#358 Search tasks by args/kwargs

Melevir:


mher:
Thanks for the pull request! I'd suggest a couple of improvements:

1. Use search language like github search
    `foo`  find all tasks containing foo in args, kwargs or result
    `args:foo` find all tasks containing foo in arguments
    `kwargs:foo=bar` find all tasks containing foo=bar keyword
    `result:foo` find all tasks containing foo in result

    if the search term contains spaces it should be enclosed in “ (e.g. args:”hello world”)
2. Document search behavior
3. Move Search, Received and Started controls to the second column of filter box

Melevir:
If I missed something or you have more improvements - please, tell me. I'll be glad to fix that.

mher:
Sorry for the late answer. It looks like there are some problems with column alignments.
untitled

We can try to narrow the space between columns and/or make Received/Started boxes smaller.

mher:
Merged, thanks! Will fix the alignment later.

Melevir:
Damn, totally forgot about that. Sorry, will fix that.

PR #365, opened 2015-04-01
Wait results option

xBeAsTx:
So, sometime it's usefull send task and get results in same http-query. Moreover, it allows not to call every 1-2-3 seconds for requesting results by '/api/task/result/' method.

This pull request implements it.

Example:
```
curl -X POST -d '{"args":["..."]}' http://127.0.0.1:5555/api/task/apply/task.foo
```


flower/api/tasks.py
```diff
@@ -113,6 +119,10 @@ def post(self, taskname):
 
         result = task.apply_async(args=args, kwargs=kwargs, **options)
         response = {'task-id': result.task_id}
+        if 'wait' in options:
+            # wait until task finished and do not raise anything
+            result.get(propagate=False)
```
mher: result.get will block the event loop and other requests couldn't be handled

xBeAsTx:
Now it's asynchronously

flower/api/tasks.py
```diff
@@ -113,9 +120,25 @@ def post(self, taskname):
...
+
+    def async_wait(self, result, response):
+        from threading import Thread
+        def wait_results(result, response):
+            # wait until task finished and do not raise anything
+            r = result.get(propagate=False)
+            self.update_response_result(response, result)
+            self.write_result(result, response)
+        th = Thread(target=wait_results, args=(result, response, ))
+        th.start()
```
mher: th.join() is missing here

mher:
Can you move this functionality into a separate handler and document it? /api/task/apply/(.+) seems a good candidate.

mher:
Merged, thanks!

PR #380, opened 2015-04-14
Document the environment variables that OAuth reads.

jibsheet:
While there is code in command.py to read FLOWER_FOO and turn that into
options.foo, the oauth code looks for oauth2_key (parsed from a config
file) FLOWER_GOOGLE_OAUTH2_KEY will never be found since it will be
placed in options.google_oauth2_key.

I've just fixed the docs to match the code, it may be preferable to fix
the code to match the docs (while maintaining backcompat).


mher:
Thanks for the pull request. I've modified the docs.

Environment variables should be prefixed with FLOWER_ without GOOGLE_ prefix (e.g. FLOWER_OAUTH2_KEY). For backward comparability FLOWER_GOOGLE_ will also work. As documented before.

PR #389, opened 2015-04-28
Don't use lstrip when removing ENV_VAR_PREFIX in the flower command

mdallen22:
Don't use lstrip when removing ENV\_VAR\_PREFIX in the flower command. lstrip removes all occurrences of the given characters at the beginning of a string, it does not treat these characters as a substring. We need to remove the substring "FLOWER_", not the characters "F", "L", "O", "W", "E", "R", "\_". For example, given "FLOWER\_", lstip will remove "FLOWER\_O" from "FLOWER\_OAUTH2\_SECRET", but only "FLOWER\_" should be removed.

### DIFF
--- a/flower/command.py
+++ b/flower/command.py
@@ -32,7 +32,7 @@ class FlowerCommand(Command):
         env_options = filter(lambda x: x.startswith(self.ENV_VAR_PREFIX),
                              os.environ)
         for env_var_name in env_options:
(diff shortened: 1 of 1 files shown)


mher:

kneckinator:
What was the reasoning behind the removal of ENV_VAR_PREFIX? I have tried my best googling this but have yet to find anything tangible. Was this related to security?

PR #396, opened 2015-05-19
Allow other (external) login handlers

pferreir:
I guess we're probably not the only ones wanting to integrate Flower authentication with our own app's. The current Google-only OAuth2 mechanism doesn't help in that regard.

This patch allows authentication to be performed by a different Python package. It is the simples solution I could find. Maybe not the optimal one, but I'm available to improve it if needed.

Said package would have to specify the following in its `setup.py`:

```python
# [...]
    entry_points={'flower.auth_providers': {'login_handler = another_package:GithubLoginHandler'}}
# [...]
```

### DIFF
--- a/flower/views/auth.py
+++ b/flower/views/auth.py
@@ -16,6 +16,8 @@ from ..views import BaseHandler
 
 
 class GoogleAuth2LoginHandler(BaseHandler, tornado.auth.GoogleOAuth2Mixin):
+    _OAUTH_SETTINGS_KEY = 'oauth'
+
     @tornado.web.asynchronous
     def get(self):
         redirect_uri = self.settings[self._OAUTH_SETTINGS_KEY]['redirect_uri']
@@ -65,7 +67,7 @@ class GithubLoginHandler(BaseHandler, tornado.auth.OAuth2Mixin):
     _OAUTH_AUTHORIZE_URL = "https://github.com/login/oauth/authorize"
     _OAUTH_ACCESS_TOKEN_URL = "https://github.com/login/oauth/access_token"
     _OAUTH_NO_CALLBACKS = False
-    _OAUTH_SETTINGS_KEY = 'google_oauth'
+    _OAUTH_SETTINGS_KEY = 'oauth'
 
     @tornado.auth._auth_return_future
     def get_authenticated_user(self, redirect_uri, code, callback):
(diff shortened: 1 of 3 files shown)


pferreir:
Example external login handler: https://gist.github.com/pferreir/2b1b0cd33c95631ce229

pferreir:
If this goes forward, it would be great to have the OAuth options under a more general-looking setting, instead of `google_oauth`.

mher:
Thanks for the pull request! I've added `auth_provider` option  https://github.com/mher/flower/commit/1d8bf46a1161a1b5709ca4f404f872991b60e466 to allow custom authentication methods

mher:
Can you add your github auth handler to auth.py?

mher:
@pferreir `google_oauth` is already renamed in the master

ThiefMaster:
The new auth_provider setting works fine. However, a way to hook into the `BaseHandler.get_current_user` logic would be greatly appreciated. Right now we need an ugly hack if we don't want email-based authentication (we ask the OAuth provider if the user is an administrator and in that case grant him access). Right now we use `--auth='^Admin$'` and set this as the username after login. Not very pretty...

pferreir:
Done. I added a second commit that removes the "Google" branding and replaces it with more neutral variable names. I think it makes sense, since `LoginHandler`s will want to use those command-line options and having them be `GOOGLE_*` would confuse people.

flower/views/auth.py
```diff
@@ -58,6 +62,85 @@ def __new__(cls, *args, **kwargs):
         return instantiate(options.auth_provider, *args, **kwargs)
 
 
+class GithubLoginHandler(BaseHandler, tornado.auth.OAuth2Mixin):
+
+    _OAUTH_AUTHORIZE_URL = "https://github.com/login/oauth/authorize"  # "https://oauth.web.cern.ch/OAuth/Authorize"
```
ThiefMaster: you forgot to remove the cern oauth urls here

pferreir: Thanks! I think comments are becoming invisible to my eyes...

mher:
Thanks @pferreir ! Can you update the docs as well? 

https://github.com/mher/flower/blob/master/docs/auth.rst

pferreir:
Still haven't had the time to work on the docs, however :wink:

Issue #387, opened 2015-04-24
worker list shows lots of temporary clients


frasergraham:
I just updated to the latest development flower and anything that connects to the broker to create a new job seems to be adding an entry to the worker list..

gen26865@myapp-api-dev	Offline	0	0	0	0	0	N/A

I get hundreds of these in the dashboard view and it makes the view unusable for knowing if my workers are online.

frasergraham:
I am using celery 3.1.11 and RabbitMQ3.4 if that makes a difference

saich:
I'm having the same issue. Using flower==0.8.2, celery==3.1.17, redis==2.10.1.

frasergraham:
I upgraded my celery install to 3.1.18 and I still see this, i'm using flower from master. Could this be related to how i've configured django? It still uses django-celery so it never defines an app?

mher:
Fixed by #417

PR #417, opened 2015-06-22
dashboard: do not show temporary clients

westphahl:
Starting with Flower 0.8.1 the dashboard started showing offline worker entries
for every client.

This issue is fixed by filtering out all non-worker entries.

Fixes: #387


westphahl:
Please let me know if there is anything that needs to be changed.

westphahl:
I'm a little disappointed that this pull request is pending for so long without any comments. Especially since it's fixing an open issue (#387).

mher:
Merged, thanks! Sorry for late response.

PR #420, opened 2015-06-30
bugfix: remove deprectated url_prefix method in ajax call

TimVaes:
Fixed js runtime error below:
image


mher:

PR #422, opened 2015-07-02
Update Dockerfile

keeper575:
pip could not be imported in some OS. Hence a separate pip install will remove the problem.


mher:

PR #432, opened 2015-07-18
returning temporary result during task progress

asmodehn:
Conflicts:
	flower/api/tasks.py

This is a cherry pick from the change I tested on 0.8. untested after this conflict resolution.

The point here is that a celery task can return temporary results ( via [update_state()](http://celery.readthedocs.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.update_state) and maybe passing a custom state ) with some data in result, and that it is useful to be able to retrieve this data while the task is still ongoing... => we do not need to wait for ready().

Let me know if I should make more changes...

### DIFF
--- a/flower/api/tasks.py
+++ b/flower/api/tasks.py
@@ -307,8 +307,8 @@ Get a task result
         if not self.backend_configured(result):
             raise HTTPError(503)
         response = {'task-id': taskid, 'state': result.state}
-        if result.ready():
-            self.update_response_result(response, result)
+
+        self.update_response_result(response, result)
         self.write(response)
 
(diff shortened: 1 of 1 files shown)


mher:
Thanks for the pull request! Waiting for the result is still useful. It would be better to add some `no-wait` parameter the the handler.

As an example you can have a look at https://github.com/mher/flower/blob/master/flower/api/tasks.py#L419

asmodehn:
I also realized that the display in webpage should also probably be changed.
Now if the task is in progress, the web interface shows `result: None`, even if a result is already set...

Issue #439, opened 2015-08-01
missing broker stats for redis


jerzyk:
when using redis with priority queues, flower is displaying only requests that are assigned only with queue and priority 0.

Prioritization is done internally by kombu's redis transport (kombu/transport/redis.py:Channel).
Kombu is creating n queues (controlled by `PRIORITY_STEPS` parameter, default: [0, 3, 6, 9]).
Queue names are created by concatenating queue name, prefix (default: '\x06\x16') and priority for all priority values except 0, for 0 (highest) a raw queue name is used.

Priority steps were implemented in version 3.0, it's default value can be override by specifying `priority_steps` as array in sort order in `BROKER_TRANSPORT_OPTIONS` config variable.

Flower does not read celery configuration nor have an option to specify what queue names shall be inspected. Instead all queue names are read by inspecting workers. Unfortunately in this scenario, only the raw queue name is reported by workers, not their prioritized names.

I can try to add this functionality, but there are design decisions needed. Basic implementation can be done in three ways:
1. add option to specify all queue names - I do not like this, because it looks like step backward, but it may be used as "additional queues to look at" - but then there will be multiple queues on the graph (not necessary bad thing)
2. add option to specify redis priorities only - much better, because then, `llen` will be used by inspecting all workers active queues, plus it is quite simple to implement
3. extend option 2 to read all default config from celery - definitely more work 

Let me know which option or any other is better, so I can help.

mher:
Actually readnig celery options is pretty trivial. Try `self.capp.conf.BROKER_TRANSPORT_OPTIONS`. Those options can be passed to https://github.com/mher/flower/blob/master/flower/utils/broker.py#L99 and it will allow to avoid introducing a new flower option.

PR #440, opened 2015-08-01
Redis priority queues

jerzyk:
as discussed in #439

### DIFF
--- a/flower/views/broker.py
+++ b/flower/views/broker.py
@@ -24,8 +24,12 @@ class BrokerView(BaseHandler):
             http_api = app.options.broker_api
 
         broker = Broker(app.capp.connection().as_uri(include_password=True),
-                        http_api=http_api)
+                        http_api=http_api, celery_conf=self.capp.conf)
         queue_names = ControlHandler.get_active_queue_names()
+
+        if not queue_names:
+            queue_names = {self.capp.conf.CELERY_DEFAULT_QUEUE}
+
         queues = yield broker.queues(sorted(queue_names))
 
         self.render("broker.html",
--- a/flower/views/broker.py
+++ b/flower/views/broker.py
@@ -18,13 +18,15 @@ class BrokerView(BaseHandler):
     @gen.coroutine
     def get(self):
         app = self.application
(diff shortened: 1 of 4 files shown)


flower/views/broker.py
```diff
@@ -24,8 +24,12 @@ def get(self):
             http_api = app.options.broker_api
 
         broker = Broker(app.capp.connection().as_uri(include_password=True),
-                        http_api=http_api)
+                        http_api=http_api, celery_conf=self.capp.conf)
         queue_names = ControlHandler.get_active_queue_names()
```
mher: I think it would be better to pass only BROKER_TRANSPORT_OPTIONS. Broker should get only broker specific options.

flower/views/broker.py
```diff
@@ -24,8 +24,12 @@ def get(self):
...
+
+        if not queue_names:
+            queue_names = set([self.capp.conf.CELERY_DEFAULT_QUEUE])
+
```
mher: :+1:

requirements/default.txt
```diff
@@ -2,3 +2,4 @@ celery>=3.1.0
 tornado>=4.0.0
 babel
 pytz
+futures>=3.0.3
```
mher: futures is required only for python2. It is added to requirements here https://github.com/mher/flower/blob/master/setup.py#L48

flower/views/broker.py
mher: Broker is also instantiated in /api/tasks.py

jerzyk: there are so many other options that could be used, so idea was to pass all of them once, to have them ready if some other options will be needed (e.g. for mongodb broker, those options are under different name CELERY_MONGODB_BACKEND_SETTINGS)

requirements/default.txt
jerzyk: removed

flower/views/broker.py
jerzyk: changed all, as you wanted.

mher:

PR #441, opened 2015-08-02
move docs and code menu items to the right

jerzyk:
it is less confusing for the users, especially that those two links lead to external addresses - which is important for organizations with limited internet access


mher:
Merged, thanks!

PR #444, opened 2015-08-07
0.5

pygeek:


pygeek:
PS: .5 branch didn't pass to begin with.

pygeek:
Also, note that without this fix, .5 build fails on install.

mher:
Version 0.5 is released in 2013. Unfortunately, I don't have a time to release a bugfix for it. You can install the dependencies manually or uptate to the latest version.

Issue #445, opened 2015-08-11
/api/tasks broken in versions > 0.8.2


inactivist:
Hitting /api/tasks results in an empty response (or blank page if you're using a browser), and an exception message on the console:

    [E 150811 13:53:00 web:1407] Uncaught exception GET /api/tasks (127.0.0.1)
        HTTPServerRequest(protocol='http', host='localhost:5555', method='GET', uri='/api/tasks', version='HTTP/1.1', remote_ip='127.0.0.1', headers={'Accept-Language': 'en-US,en;q=0.8', 'Accept-Encoding': 'gzip, deflate, sdch', 'Host': 'localhost:5555', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8', 'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.130 Safari/537.36', 'Connection': 'keep-alive', 'Cookie': '<snip>', 'Cache-Control': 'max-age=0', 'Upgrade-Insecure-Requests': '1'})
        Traceback (most recent call last):
          File "/usr/local/lib/python2.7/site-packages/tornado/web.py", line 1332, in _execute
            result = method(*self.path_args, **self.path_kwargs)
          File "/usr/local/lib/python2.7/site-packages/tornado/web.py", line 2607, in wrapper
            return method(self, *args, **kwargs)
          File "/Users/someuser/.virtualenvs/someproject/lib/python2.7/site-packages/flower/api/tasks.py", line 447, in get
            worker=worker, state=state):
          File "/Users/someuser/.virtualenvs/someproject/lib/python2.7/site-packages/flower/utils/tasks.py", line 21, in iter_tasks
            any_value_search_term = search_terms.get('any', None)
        AttributeError: 'NoneType' object has no attribute 'get'

Looking at the [tasks api code](https://github.com/mher/flower/blob/master/flower/api/tasks.py#L445) it appears that it's not passing search_terms, and [utils.iter_tasks](https://github.com/mher/flower/blob/master/flower/utils/tasks.py#L25) requires that parameter (compare [views/tasks usage](https://github.com/mher/flower/blob/master/flower/views/tasks.py#L59)) -- I've not debugged further, but it appears `utils.iter_tasks` requires a dict-like `search_terms` parameter (though it defaults to `None`).

Additional info:

OSX Mavericks
[...]

PR #460, opened 2015-09-04
Extracted search terms safely to fix /api/tasks

alexmojaki:
Also note that `args_search_terms` was defined twice. By the way `dict.get` always defaults to `None` (https://docs.python.org/2/library/stdtypes.html#dict.get).


alexmojaki:
Fixes #445

mher:
Fixed by https://github.com/mher/flower/commit/79ef3fe160322781f567bbf88a3289803d1fe2e7

PR #466, opened 2015-09-14
Make sure Babel is at least 1.0

danilotsr:
`babel.dates.format_timedelta` is supported only on babel-1.0 onward, see their [release notes](http://babel.pocoo.org/docs/changelog/#version-1-0).

Besides, updated a test to use `assert_called_once_with`, given that `assert_called_once` is no longer supported in [Python 3's mock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock).

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,4 +1,4 @@
 celery>=3.1.0
 tornado>=4.0.0
-babel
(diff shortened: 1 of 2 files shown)


mher:

PR #472, opened 2015-09-21
Updating readme for OAuth

lucas:
Updating readme to be more verbose around Google OAuth requirements, since Google+ API must be enabled, and it is not mentioned in the docs currently.


mher:

PR #479, opened 2015-10-08
Update AUTHORS

lucas:
Adding self for docs update


AUTHORS
```diff
@@ -58,3 +58,4 @@ Gaurav Kumar
 Simon Westphahl
 Pedro Ferreira
 Danilo Resende
+@lucas
```
mher: This file is only for real names

PR #483, opened 2015-10-17
Add back support for --url_prefix

theY4Kman:
This PR adds back the `--url_prefix` option. It modifies the routes to include the URL prefix, so a proper nginx config would look simply like:

```nginx
server {
    # ...
    location /flower/ {
        # no "rewrite /flower/(.*) /$1 break;" line
        proxy_pass http://localhost:5555;
    }
}
```

And run flower with:

```bash
flower --url_prefix=/flower/
```

### DIFF
--- a/flower/app.py
+++ b/flower/app.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import logging
+import re
 
 from functools import partial
 from concurrent.futures import ThreadPoolExecutor
@@ -11,7 +12,7 @@ import tornado.web
 from tornado import ioloop
 
 from .api import control
-from .urls import handlers
+from .urls import get_handlers
 from .events import Events
 from .options import default_options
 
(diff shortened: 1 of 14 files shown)


ranman:
I'd love to see this or #508 merged

theY4Kman:
It doesn't seem like this will be merged. You might be okay with packaging up a version of Flower containing the code in this PR. I would recommend using this PR, because it not only changes the URL handlers, but also accounts for URLs in the JavaScript, so real-time components like the monitoring graphs will work.

mher:
I reimplemented the support of --url-prefix option. Please try the master version.

mher:
@theY4Kman thanks for the pull request! It is better to use `reverse_url`.

theY4Kman:
I'm glad to hear it! I'll take a look at it later; I'd love to learn a
better way. My proposition felt a bit hackish.

emaniacs:
I use flower with --url_prefix=flower and proxy with nginx so the flower url is http://mydomain/flower/,
but dashboard never updated(Websocket) because of condition in file https://github.com/mher/flower/blob/master/flower/static/js/flower.js#L560

`560: if ($.inArray($(location).attr('pathname'), ['/', '/dashboard']) !== -1) {`

I think you need expose url_prefix value in template.

PR #485, opened 2015-10-22
Dashboard workers display is now sorted by worker's name

wavenator:
It is hard to look at the workers table as it is not sorted in any intuitive order.
This is a quick patch to make some order.
I'll work on some more comprehensive at the future.


mher:
Thanks for the PL. Is it possible to hide/disable pagination?

mher:
Merged, thanks!

Issue #499, opened 2015-12-14
collectstatic failure when using ManifestStaticFilesStorage (Django 1.7+)


lisac:
I'm using flower 0.8.3 in a Django 1.8.2 project.

When I set my Django project to use the [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/1.8/ref/contrib/staticfiles/#manifeststaticfilesstorage) option (introduced in Django 1.7), my `collectstatic` step fails because of file paths referenced in the [jquery-ui.css](https://github.com/mher/flower/blob/master/flower/static/css/jquery-ui.css) that are not accounted for in the repository, eg ` url(images/ui-bg_flat_75_ffffff_40x100.png)` and `url(images/ui-bg_glass_95_fef1ec_1x400.png)`.
 
This is the error I get from `python manage.py collectstatic`:
```
[...]
Post-processed 'css\bootstrap.min.css' as 'css/bootstrap.min.129a5e00854b.css'
Post-processed 'css\flower.css' as 'css/flower.c956fd18aac1.css'
Post-processing 'css\jquery-ui.css' failed!

Traceback (most recent call last):
  File "manage.py", line 7, in <module>
    execute_from_command_line(sys.argv)
  File "C:\Projects\athena\.venv\lib\site-packages\django\core\management\__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "C:\Projects\athena\.venv\lib\site-packages\django\core\management\__init__.py", line 330, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "C:\Projects\athena\.venv\lib\site-packages\django\core\management\base.py", line 390, in run_from_argv
    self.execute(*args, **cmd_options)
  File "C:\Projects\athena\.venv\lib\site-packages\django\core\management\base.py", line 441, in execute
    output = self.handle(*args, **options)
  File "C:\Projects\athena\.venv\lib\site-packages\django\contrib\staticfiles\management\commands\collectstatic.py", line 168, in handle
    collected = self.collect()
  File "C:\Projects\athena\.venv\lib\site-packages\django\contrib\staticfiles\management\commands\collectstatic.py", line 120, in collect
    raise processed
ValueError: The file 'css/images/ui-bg_flat_75_ffffff_40x100.png' could not be found with <django.contrib.staticfiles.storage.ManifestStaticFilesStorage object at 0x036B7690>.
```

PR #500, opened 2015-12-14
fix issue #499

lisac:
Add files for the image filepaths referenced in [jquery-ui.css](https://github.com/mher/flower/blob/master/flower/static/css/jquery-ui.css).

Fixes https://github.com/mher/flower/issues/499.


mher:
Merged thanks!

PR #505, opened 2015-12-29
Add received and started in API list tasks

raimonesteve:
When working with API, received and started parameters filter tasks by dates.

Example:

``` html
http://localhost:5555/api/tasks?received-start=2015-12-28 23:00
http://localhost:5555/api/tasks?received-start=2015-12-28 23:00&received-end=2015-12-29 23:00
```

Also you could search a task by UUID (if exist or not) without call request info uri.

Example:

``` html
http://localhost:5555/api/tasks?task-id=ada92cb7-8c4e-4b29-8f63-05660ccc3ad1
```

### DIFF
--- a/flower/api/tasks.py
+++ b/flower/api/tasks.py
@@ -435,6 +435,10 @@ List tasks
         worker = self.get_argument('workername', None)
         type = self.get_argument('taskname', None)
         state = self.get_argument('state', None)
+        received_start = self.get_argument('received-start', None)
+        received_end = self.get_argument('received-end', None)
+        started_start = self.get_argument('started-start', None)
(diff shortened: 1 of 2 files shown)


mher:
Why don't you use /task/info api method? It returns task by UUID.

raimonesteve:
Why to do a lot of API connection to know each UUID task? For example, to do a list/view tasks (or search/read tasks).

jheld:
What's the status of this?

jheld:
I think this is worthwhile. I think I did something similar on my Elasticsearch PR.

Can we add a test and consider a merge @mher?

PR #508, opened 2016-01-06
enable --url-prefix option

aleiphoenix:
Any other convenient way to prefix Tornado app?

I need to prefix flower app to a certain path ,  say `http://example.com/flower`.


aleiphoenix:
I've noticed that all url in templates are hard coded. So `reverse_url` is needed.

ranman:
+1

mher:
Please try the master version. It already supports --url-prefix option.

PR #510, opened 2016-01-06
use timezone from celery settings if configured

vinayinvicible:
Currently datetime values in tasks view are displayed using UTC as timezone. And also the timezone info is not being shown with the values.

Modified the code to display datetime values using the timezone configured in celery settings. If timezone is not configured, timezone will default to UTC. Also display the timezone info along with the value.


mher:
Merged, thanks!

vinayinvicible:
May I know when will the new version be released.

PR #513, opened 2016-01-12
add option `unix_socket` for flower

wklken:
Deploy on docker and expose the socket file. It's simpler, and tornado support that.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,4 +59,4 @@ Gaurav Kumar
 Simon Westphahl
 Pedro Ferreira
 Danilo Resende
-wklken
+Kevin Wu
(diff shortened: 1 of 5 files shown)


AUTHORS
```diff
@@ -59,3 +59,4 @@ Gaurav Kumar
 Simon Westphahl
 Pedro Ferreira
 Danilo Resende
+wklken
```
mher: Only real names should be listed. Github handles can be extracted from git logs.

wklken:
@mher I just change the name:0)

Flower is nice. Now I am deploying flower in docker, with celery 3.1.6, And run into some bugs.

in the requirements

```
celery>=3.1.0
```

But some code in new patch use the new methods in later version celery(3.1.19).
I will make a pr to fix that later.

mher:
Merged, thanks!

PR #520, opened 2016-01-22
Dashboard group worker by hostname

PMickael:
I made this little update for my own usage, maybe someone will like it so, I add it to pull request.

flower

Mickael

### DIFF
--- a/flower/static/css/flower.css
+++ b/flower/static/css/flower.css
@@ -229,3 +229,7 @@
   width: 100%;
   display: inline-block;
 }
+
+tbody th {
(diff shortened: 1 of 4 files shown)


mher:
@PMickael I've refactored the dashboard. It is now possible to filter workers.

PMickael:
@mher not useful for my usage, I have too much worker and I need to know by host which worker have a problem, and If every core of every server is in use, so 8 or 16 worker by host

johnarnold:
IMO, cool feature if you have a lot of workers per host, but perhaps you can make it optional / checkbox?

PR #525, opened 2016-02-05
Task abort implementation, rebased onto master

asmodehn:


asmodehn:
master is not working for me (CORS patches missing among other stuff), so I tested this on my branch before rebasing.

Let me know if I should do more before we can merge this. thanks !

flower/api/control.py
```diff
@@ -471,6 +471,44 @@ def post(self, taskid):
         self.write(dict(message="Revoked '%s'" % taskid))
 
 
+class TaskAbort(ControlHandler):
```
mher: Looks like this class is not used anywhere

mher:
Merged, thanks!

PR #533, opened 2016-02-25
Fallback when app have no attribute 'timezone'

bmwant21:
* Prevent failures when Celery object has no attribute `timezone`


mher:

PR #535, opened 2016-02-29
Add option to set unix_socket permissions

ba1dr:
By default Tornado creates a socket file with 600 permission flags. But running behind of nginx there is an "Access Denied" error as nginx might work under different user.
I am going to use it with 666 permission flag and hope it would be useful for others.

### DIFF
--- a/flower/app.py
+++ b/flower/app.py
@@ -52,7 +52,8 @@ class Flower(tornado.web.Application):
         else:
             from tornado.netutil import bind_unix_socket
             server = HTTPServer(self)
-            socket = bind_unix_socket(self.options.unix_socket)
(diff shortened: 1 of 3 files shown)


ba1dr:
Actually there should be 2 separate pull requests, but they got merged into one..
Second commit helps to extend the Flower by subclassing it's FlowerCommand and Flower classes with adding more handlers for example.

For example I can create a basic app with a new API handler - see example: https://gist.github.com/ba1dr/5524797cd84b19706d19

mher:
I'd avoid creating a special option for unix socket mode. I think it makes sense to change the default to 666.

mher:
Also, please send a separate PL for different changes. Thanks!

PR #536, opened 2016-03-04
Add max_workers option

themylogin:
Add max_workers option. This is required when running celery workers in docker containers because they get new unique hostname on each deploy and flower is getting flooded with offline workers from stopped containers. Specifying --max-workers=10 solves this problem.


mher:

PR #538, opened 2016-03-04
replaced map to list comprehension

mastak:
In python 3 `map` returns generator and tornado raise exception while processing env variable like:
`FLOWER_BASIC_AUTH="admin:passw1,user:passw2"`

```
tornado.options.Error: Option 'basic_auth' is required to be a list of str
```


mher:
Merged, thanks!

PR #539, opened 2016-03-06
extract_settings() require previous settings as well

kobim:
Some settings require previous values (e.g. `login_url` and `url_prefix`), hence instead of updating the settings dictionary, override it in-place.


mher:
Merged, thanks!

PR #540, opened 2016-03-07
Fix stale cache for workers

prmtl:
If any of the inspect methods for workers returns "falsy" results
(like empty array, which is completely valid response from celery)
cache for workers info is not updated which causes inaccurate results
returned by the API.


mher:

prmtl:
Cool, will do!

BTW @mher, when do you plan to release a new version?

mher:
In a week or two

Issue #543, opened 2016-03-09
Feature request: Tasks to be updated via Websockets


AndreLobato:
When monitoring is needed to keep hitting refresh to see new info about tasks in the Tasks list, I wonder if have plans to add automatic update pushed via Websockets. I had implement this in the past in a fork of version 0.7.3, using the events API, but one of the problems I had was the fact we have 1 address for each task-state, that means I had to connect several clients instances, one for each state. 

Would worth to work in a PR for that?

mher:
Feel free to send a pull request

mher:
I'm not sure using web sockets is the best solution. I think Datatable's auto update is more desirable.

PR #552, opened 2016-03-20
Added tasks table update via WebSockets

AndreLobato:
Implement DataTable updates via Websocket push #543 

Method may be incomplete still, but works fine, I have not yet run the test suite.

Suggestion for improvement, to add a switcher button to turn on/off the auto-updating since application could be processing too many tasks, making impossible to read the table.

### DIFF
--- a/flower/api/events.py
+++ b/flower/api/events.py
@@ -20,7 +20,6 @@ class EventsApiHandler(BaseWebSocketHandler):
 EVENTS = ('task-sent', 'task-received', 'task-started', 'task-succeeded',
           'task-failed', 'task-revoked', 'task-retried', 'task-custom')
 
-
 def getClassName(eventname):
     return ''.join(map(lambda x: x[0].upper() + x[1:], eventname.split('-')))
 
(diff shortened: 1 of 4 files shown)


AndreLobato:
One problem I've identified is that if you have lots of tasking running and sending events, the page would take long to load since, the draw would be called many times before the page load. so would be nice if the websocket would be connected only after the table has been drawn for the first time.

AndreLobato:
On this last update I have rationalised the calls for table.draw(), that's because, on the first implementation (calling table.draw for every new task event), at some point, the page would start to do excessive calls (GET) and jam the flower main process with too much calls, also is not a good approach since does little usage if the data sent by the WS pipe. To solve this problem I just added a time counter and set the a time threshold of 10s between each table.draw call. Also it will only try to call draw on task-received events, on all others events would just update the table row data if present, making a better usage of data flowing in the WS protocol.

alcidesv:
Will this PR still allow Flower to run behind an HTTP/2 reverse proxy?

mher:
WebSockets were removed from dashboard and I'm going to deprecate the API implementation. It is better to implement this using Datatables auto refresh.

Issue #356, opened 2015-03-27
Feture request: Time window in monitor


MasterMind2k:
Hi.

Currently when monitoring, all points get shown. It would be nice to specify the last x seconds|minutes|hours monitoring windows.

Best regards.

mher:
Thanks for the feedback!

mher:
Monitor tab is deprecated, use Prometheus instead. It supports the requested feature.

PR #560, opened 2016-04-03
Implement time window ability to dashboard, fix responsive issues, and more improvements

hakanw:
Hello,

I've been using this weekend to make some improvements to the dashboard. To be able to make these charts work better I've changed the use of rickshaw into Highcharts instead. It works very nicely and is ready for further improvements! 

Here's a summary of the changes:

* Implement time window / rolling time window support, #356 
* Add markers and units to both X and Y axis to make graphs more clear
* Fix bugs with responsiveness, so it works correctly with bootstrap without adding scrollbars now

Hope you like it!

I would be keen to add the ability for the dashboard page to pull in older data (if it exists in memory by the worker) so it doesn't disappear each time you change tab too.


hakanw:
Here's an example how it looks. 

* We've got the time window dropdown in the upper right like on AWS, and the graphs change what they're using dynamically based on this option.
* You can also see the labels on the X axis which are dynamically changed based on how long span you're looking at. 
* And "Tasks" is written on the Y axis. This is "Seconds" where time is being measured.

skarmavbild 2016-04-03 kl 16 47 00

mher:
I don't think we can use HighCharts in Flower. HighCharts is only free for personal use and non-profit organizations.

hakanw:
According to this stackoverflow question it should be OK for use in open source:

http://stackoverflow.com/questions/16132822/include-highcharts-in-open-source-project

I'll dig more into it.

hakanw:
Reading on the FAQ.. https://shop.highsoft.com/faq  It's a bit unclear. It says you can use it for open source, and redistribute given these two conditions:

```
You are allowed to distribute a Highsoft software product with non-commercial packages given that you fulfill two conditions:

- Emphasize to your users that Highsoft software products are not free for commercial and Governmental use. You can do this on your download page or when your users activate a Highsoft Software Product in your application. 
- Provide a link back to highcharts.com in the same location.
```

mher:
The licence allows to include it in an open source project but the project users should pay for it. I think it is not acceptable for flower.

PR #563, opened 2016-04-09
Fix monitor tab problem of missing graphs

jzhou77:
When url_prefix is enabled, the Monitor tab misses Ajax update because URLs do
not have the url_prefix. Fix by adding the prefix to these Ajax calls.


mher:

PR #574, opened 2016-05-10
Use Alpine as Docker base image

tnir:
- Use the latest OS
- Minimize Docker image size
- Fix flower version
- Update docs for Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,5 @@
-FROM ubuntu:12.04
+FROM python:3.4-alpine
 
-RUN apt-get update
-RUN apt-get install -y wget ca-certificates
+ENV FLOWER_VERSION 0.9.1
 
-# Install pip
-RUN cd /tmp; wget https://raw.github.com/pypa/pip/master/contrib/get-pip.py;\
- python get-pip.py; rm get-pip.py;
-
-# Install pip via package-manager 
-RUN apt-get install -y python-pip
-
-# add flower
-RUN pip install redis flower
-
+RUN pip install redis flower==$FLOWER_VERSION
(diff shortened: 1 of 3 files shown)


tnir:
@mher Could you have time to review this MR as well as unreviewed PRs? 😄

Dockerfile
```diff
@@ -1,15 +1,5 @@
-FROM ubuntu:12.04
+FROM python:3.4-alpine
```
mher: I'd prefer to have a more general distribution. No need to limit to python 3.4

Dockerfile
```diff
@@ -1,15 +1,5 @@
...
-# Install pip
-RUN cd /tmp; wget https://raw.github.com/pypa/pip/master/contrib/get-pip.py;\
- python get-pip.py; rm get-pip.py;
-
-# Install pip via package-manager 
-RUN apt-get install -y python-pip
-
-# add flower
-RUN pip install redis flower
+ENV FLOWER_VERSION 0.9.1
```
mher: `pip install -U flower` installs the latest version. No need to specify the version.

Dockerfile
tnir: Yes, I'd like to use popular distro too.

OK, then, why don't we use xenial (Ubuntu 16.04 LTS)? https://github.com/tnir/flower/blob/docker-xenial/Dockerfile

As the image from xenial has as big as 212MB, it's inconvenient for everyone to use it. Note that alpine-based flower image has just 49MB.

I'm providing the images on https://hub.docker.com/r/tnir/flower/tags/.

Dockerfile
tnir: How can we specify the flower version, then?

jvrplmlmn:
```
FROM python:3.4-alpine

ENV FLOWER_VERSION latest

RUN if [ "${FLOWER_VERSION}" != "latest" ]; then pip install redis flower==${FLOWER_VERSION}; else pip install redis flower; fi
```

mher:
Fixed by #644

PR #583, opened 2016-05-26
Fix a bug in humanizing timestamps on the tasks page

SharangP:
Previously, these were not humanized as timestamps with timezone info and were inconsistent with the timestamp displayed on the task/ page


mher:

PR #620, opened 2016-10-05
Monitor Tab not showing tasks

### LINKED ISSUE #568: Monitor Tab not showing tasks
Investigating the issue I found that this is a problem with how flower manages to get time. Here is what I think should be changed:

The generation of the _lastquery_ timestamp is coming from javascript, that means that if the user computer has a problem on his clock he will not be able to see tasks coming and others will (making the bug hunting a pain). This could be avoided by using python's **time** library and adding a _padding_ to that time instead which will use the server time. We could also add a url query like _lastquery_ to modify the padding if required.

Here is a code example:

```
time_padding = self.get_argument('padding', type=float)
timestamp = time.time() - padding # let's say padding is 60 seconds
state = self.application.events.state

data = defaultdict(int)
for _, task in state.itertasks():
    if (timestamp < task.timestamp and task.state == states.SUCCESS):
        data[task.worker.hostname] += 1
```

Also, to be honest, I don't quite understand why these timestamp comparisons on the monitor view even exists, for what I could see, the tasks are sent to the monitor just once, even if the times are not exactly matching, shouldn't we just show them?

I understand this may be the case with RabbitMQ but not with other brokers and that would make sense. I think a padding to the time would make things a lot easier to handle though.

These are my installed app versions:
flower==0.9.1
celery==3.1.23

gmixo:
Fix #568 
It seems timestamp comparison is wrong or even don't needed

### DIFF
--- a/flower/views/monitor.py
+++ b/flower/views/monitor.py
@@ -25,7 +25,7 @@ class SucceededTaskMonitor(BaseHandler):
 
         data = defaultdict(int)
         for _, task in state.itertasks():
-            if (timestamp < task.timestamp and task.state == states.SUCCESS):
+            if (timestamp > task.timestamp and task.state == states.SUCCESS):
                 data[task.worker.hostname] += 1
         for worker in state.workers:
             if worker not in data:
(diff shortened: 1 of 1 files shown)


johnarnold:
This looks like it's failing due to an issue with the build job, not with the commit.  Can a maintainer fix the build?

johnarnold:
@gmixo have you verified this fixes display of tasks in monitor tab?

gmixo:
this works for me

Datta1993:
Is the issue fixed ?

johnarnold:
It seems to me that flipping the conditional doesn't really serve any purpose -- better to just remove the timestamp comparison?

gmixo:
well looks like the condition is works fine. At least on latest versions - it serve displaying of actual tasks happened from last check

PR #623, opened 2016-10-07
Removed 500px height limitation

azaitsev:
Vertical scroll not fixed to 500px and now depends on your browser height.


mher:

Issue #586, opened 2016-06-07
Why does flower require exactly `tornado==4.2.0`?


mrmachine:
Instead of `tornado>=4.2.0`? This seems very restrictive. Flower doesn't know what project it will be installed into and what packages might be installed alongside it, and they might require `4.3` (which probably also works with Flower)?

This commit relaxed all the other requirements except for tornado: https://github.com/mher/flower/commit/57e87e184edbd751b6a2ab88a886ba41e2755e0f

mher:
There was a problem with version 4.3. See https://github.com/mher/flower/issues/489 for details.

mrmachine:
So 4.3 is known to be incompatible, but are future versions of tornado also going to be incompatible? If so, Flower's requirements should be `tornado<4.3`. If it's only 4.3 that is incompatible, and future versions are expected to work, it should be `tornado>=4.2,!=4.3.0`? Although if all future versions of tornado are not going to be compatible because of an API change (rather than a bug that will be fixed), perhaps Flower needs an update to be compatible?

mher:
I didn't have a time to implement a solutions which works with both 4.2 and 4.3. The future versions should be compatible with 4.3.

mher:
Feel free to send a patch

arjenfvellinga:
Tornado 4.2.0 is considered vulnerable, so we had to upgrade it in our project. http://www.tornadoweb.org/en/stable/releases/v4.2.1.html . Moreover Tornado 4..4 is already released.

Is it possible to reconsider this issue?

alzeih:
@arjenfvellinga @mher I've submitted a patch for this issue https://github.com/mher/flower/pull/629

mher:
Fixed by #629

PR #629, opened 2016-11-09
Support environment variables in tornado 4.2.0 and >=4.3

alzeih:
Tornado normalizes all option names to be dashes `-` instead of underscores `_` since 4.3.
(https://github.com/tornadoweb/tornado/commit/463669e9e085d6048e233db5a47a9e8895ff554b)

As flower handles environment variables as well, special handling is needed to support
environment variables on 4.2 and 4.3 and above.

patch for #586

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -67,3 +67,4 @@ Misha Behersky
 Sebastian Kalinowski
 Jingyu Zhou
 Maxim Krivodaev
+Alli Witheford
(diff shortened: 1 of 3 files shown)


alzeih:
PR rebased after travis errors fixed in master

sayeghr:
Definitely looking forward to this patch, any way to expedite?

mher:
Merged, thanks!

PR #630, opened 2016-11-09
fix travis build

alzeih:
drop target 2.6, not supported by celery (or dependencies including billiard)

use same targets as celery 4.0 supports, except pypy3 which doesn't work

remove "--use-mirrors", travis reports no such option


Photonios:
This should be merged, it prevents accurate reporting on the other open pull requests.

Issue #631, opened 2016-11-09
500 Error shows settings


logston:
![screen shot 2016-11-09 at 10 51 22 am](https://cloud.githubusercontent.com/assets/2641765/20144964/571d7f48-a66c-11e6-8ed4-8b555c305e23.png)

![screen shot 2016-11-09 at 10 52 07 am](https://cloud.githubusercontent.com/assets/2641765/20145059/aa6afa54-a66c-11e6-9457-2bb96f051b1c.png)

Running behind nginx. Ran `celery flower -A` to test

```
FLOWER_AUTH_PROVIDER="flower.views.auth.GithubLoginHandler"
FLOWER_BROKER_API="http://rabbituser:XXXXXXXXX@127.0.0.1:15672/api/"
FLOWER_AUTH="myemail@formygithubaccount.com"
FLOWER_OAUTH2_KEY=xxxxxxxxxxxx
FLOWER_OAUTH2_SECRET=xxxxxxxxxxxxxxxxxxx
FLOWER_AUTH2_REDIRECT_URI="https://flower.myproject.com/login"
```

Tried to test that users could not get into my flower instance without proper creds.

mher:
Fixed by #633

PR #633, opened 2016-11-09
Do not show debug information on 500 error

logston:
See https://github.com/mher/flower/issues/631 for details


mher:
Merged, thanks!

PR #634, opened 2016-11-11
The URLs would be better if configurable

antonyc:


mher:
@antonyc can you explain the use case for this change?

PR #644, opened 2016-11-26
Smaller and simpler Dockerfile.

saich:
- Smaller size (around 150MB).
- Runs on latest Python (3.5.x)
- Runs as non-root user by default.


mher:

Issue #604, opened 2016-07-29
Ability to remove offline workers


vineetgoel:
This will be super useful especially when using an Amazon ASG for tweaking the number of workers. 

Currently the dashboard looks too polluted as it shows all the offline workers too.

tbindi:
+1

zenvisuals:
+1

kheshav:
+1

clems4ever:
+1

yothinix:
Seems to be duplicate with #852

dejlek:
The difference is that #852 comes with an implementation...

mher:
https://github.com/mher/flower/pull/852 implements this

PR #649, opened 2016-12-06
Added option to hide offline workers

Photonios:
Adds a command line option that allows one to hide offline workers. This addresses #604


mher:
I'd avoid introducing a new option for hiding offline workers. It is better to hide offline workers from UI, using a URL parameter or search box filtering.

PR #655, opened 2016-12-26
Add traceback highlighting.

lmasikl:
Add `Pygments` in default dependencies.
Add new `utils.template.pygments_styles` function which return CSS for highlighting.
Traceback render raw value.
Highlight and mark task in `utils.tasks.get_task_by_id`.

### DIFF
--- a/flower/templates/task.html
+++ b/flower/templates/task.html
@@ -1,5 +1,11 @@
 {% extends "base.html" %}
 
+{% block extra_styles %}
+    <style>
+        {{ pygments_styles('.highlight') }}
+    </style>
+{% end %}
+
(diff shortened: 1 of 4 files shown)


lmasikl:
Code for issue 537

mher:
Thanks for the pull request! Traceback highlighting introduces a new dependency. I'm not sure it is worth to introduce it for very small feature.

jheld:
@mher is there another way to get this PR in, without adding an additional dependency? Or, perhaps it's worth it?

auvipy:
@mher IMHO adding pygment shouldn't be risky

PR #670, opened 2017-02-20
Fix bug in tasks page template

SharangP:
Previously, the started and received timestamps were swapped, and being
displayed in place of one another. Also, humanize the revoked timestamp.


mher:
Merged, thanks!

PR #671, opened 2017-02-27
Fix celery version comparsion

caoyi0905:
Fix celery version comparsion.

```
In [1]: '3.0.2' < '3.0.15'
Out[1]: False
```


mher:
Thanks! Merged!

PR #677, opened 2017-03-15
Fixes #618: invalid URL used for datatable query

### LINKED ISSUE #618: Invalid URL used for datatable query
Hi,

When using flower with --url_prefix, I can't list tasks (tasks tab). I see queries to `https://<url_prefix>/tasks/datatable?…`, the domain is missing. With the correct URL, the query works fine, I can get the JSON.

Using flower 0.9.1.

xavierhardy:
Fixes no graph in the monitor page too
Caused by url_prefix with a leading slash


xavierhardy:
https://github.com/mher/flower/issues/618

PR #701, opened 2017-05-31
OpenAPI 2.0 swagger spec initial commit

johnarnold:


johnarnold:
This is just an initial "static" version of a spec for flower.

I generated this spec by leveraging the apispec package, and building spec paths based on the tornado paths, as well as augmenting the Handler method docstrings with yaml where needed.  You can see this code here:
https://github.com/johnarnold/flower/tree/apispec

I think that dynamic generation is probably "generally" easier to maintain since it keeps the documentation in the context of the method, but unfortunately using this module led to duplication in the docstrings.  I did some work to trim this out using sphinx functions, but it's not finished.

I want some feedback on whether it's the right approach to imbed the swagger stuff within the docstrings.  There are a couple different swagger packages that read docstrings, so the format could be changed, but ultimately it means probably changing the docstrings to "something other than current."

mher:
The [current API](http://flower.readthedocs.io/en/latest/api.html) is generated from docstrings. What is the advantage of the suggested PL over the existing one?

johnarnold:
It's roughly a question as to whether it's worth generating the swagger spec off of the docstrings, or just hand-write the spec and keep the docstrings as-is.  Ultimately it would be nice to use python3 typing to document the params and returns, and generate the swagger spec on that -- but I haven't found a reasonable working lib for it, and that rules out 2.7 compat (unless using 2.7 static typing, which, again uses docstrings).

Given that the API is currently pretty simple, and I already wrote the initial spec, I'm ok with leaving it as a static file for now (that's what I'm running for our service), but if it sounds useful to dynamically generate, I'm happy to contribute that code too (it's written already).

johnarnold:
Can we commit the "static" file?  I'm using this in our production environment with no issues.  I'll probably add the status codes when I get some free time, but it's fully functional as is.

PR #702, opened 2017-06-01
support GitHub Enterprise

jat001:


mher:
Thanks for the pull request! The current implementation allows to extend the auth with custom implementation. You can implement GitHub Enterprise auth as a new class and enable it with `--auth_provider` option.

mikescar:
@mher  Would you consider a PR using a more consistent`FLOWER_OAUTH2_HOST` (or some such), including a separate provider class, with accompanying documentation updates?

PR #642, opened 2016-11-25
API : allow search and sort (especially by eta) and fix a unicode search in the string representation of a dict

tramora:
Hello Mher,

For our project we had to patch flower 0.8.4 (and then 0.9.0) because searching and sorting issues.
- search filter was not taken into account at all
- fix the string string and optimization
- allow sorting by eta

Would you please mind having a look please ?
Thierry

PR #705, opened 2017-06-07
Add search query filters in /api/tasks

### LINKED ISSUE #704: Allow full filtering syntax in the UI URL and in the API
The filters are quite powerful but can only be used directly from the UI. It is impossible to link a search result using a URL or to use those filters in the API (except when matching on state).

http://flower.readthedocs.io/en/latest/tasks_filter.html#task-filter-syntax
http://flower.readthedocs.io/en/latest/api.html#get--api-tasks

xavierhardy:
Fixes #704 .


xavierhardy:
Duplicated by #642 .

PR #709, opened 2017-06-21
add /api/workers/clear-cache

johnarnold:
This feature gives us a way to flush stale worker info out of the flower worker cache dictionary without restarting flower.

### DIFF
--- a/flower/api/control.py
+++ b/flower/api/control.py
@@ -23,6 +23,12 @@ class ControlHandler(BaseHandler):
         yield self.update_workers(workername=workername,
                                   app=self.application)
 
+    @gen.coroutine
+    def clear_workers_cache(self):
+        logger.debug("Clearing worker cache...")
+        self.worker_cache = {}
+        yield self.update_workers(app=self.application)
+
     @classmethod
     @gen.coroutine
     def update_workers(cls, app, workername=None):
--- a/flower/api/workers.py
+++ b/flower/api/workers.py
@@ -185,3 +185,48 @@ List workers
             self.write({workername: self.worker_cache[workername]})
         else:
             self.write(self.worker_cache)
+
+
+class ClearWorkersCache(ControlHandler):
+    @web.authenticated
+    @gen.coroutine
+    def post(self):
+        """
+Clear workers in cache
+
+**Example request**:
+
+.. sourcecode:: http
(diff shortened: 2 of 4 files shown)


mher:
I think this is not suitable for multi-user environment. It makes sense to implement the clear cache functionality as a unix signal handler. Only admins should be able to clear the server cache.

berkando:
Having an invalidation of the cache seems to be an interesting topic for others also. Is there a reason why the cache is not implemented by using some Cache-libraries? e.g. with the `cachetools.TTLCache` implementation it would be quite easy to have an automatic invalidation after a specific time.

johnarnold:
@mher there's no concept of "admins" in the API -- there's no rbac.   I don't see clear-cache being any more dangerous than shutting down a worker.  Less so, since it rebuilds immediately on the next worker inspect()  What's the concern?  It's UI-only...

mher:
@johnarnold I think it is better to add this functionality as a special parameter to `/api/workers` endpoint (see https://github.com/mher/flower/blob/master/flower/api/workers.py#L173).
The clear cache request will look like `/api/workers/?refresh=true&clear_cache=true`. What do you think?

mher:
@berkando flower uses celery's event store to keep tasks/workers. Celery store is implemented as  LRU cache https://github.com/celery/celery/blob/master/celery/events/state.py

johnarnold:
@mher that's a reasonable idea... I'll code it up and see.

jheld:
what's the status of this?

jheld:
I could see a use-case for clearing the cache for a specific worker, in the case where we know it's not gone, and we don't want to forget its information that we've stored.

Is this something we'd be interested in putting on the worker's page itself as a "only clear me" idea?

jheld:
It looks like this is a great solution for general programmatic flushing of offline workers.

We could build off of this (adding a button to the dashboard for all, or for a given worker), on a separate MR (and with a worker specific URL handler).

Once the `swagger` file is resolved of conflict, can we get this merge right in?

Cylix:
Any updates on this? That'd a great addition when integrating monitoring like datadog on top of flower

mher:
`--purge_offline_workers` option allows to filter old workers

PR #731, opened 2017-08-30
Add support for systemd socket activation

digitalpbk:
Support systemd socket activation
Example service and socket files are in `etc/`


flower/app.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import absolute_import
 
 import logging
+import os
```
auvipy: could you add test for this part?

Issue #733, opened 2017-09-12
flower can generate URLs which are too big for proxies


johnarnold:
Version: latest
Issue:  The Tasks tab generates an ajax get for /datatable?.... which can get excessively long.  On example was over 3k characters.

e.g.
[...]

PR #734, opened 2017-09-12
Update tasks datatable to use POST method

johnarnold:
This PR is to resolve https://github.com/mher/flower/issues/733

"flower can generate URLs which are too big for proxies "

### DIFF
--- a/flower/views/tasks.py
+++ b/flower/views/tasks.py
@@ -85,37 +85,7 @@ class TasksDataTable(BaseHandler):
 
     @web.authenticated
     def post(self):
-        app = self.application
-        draw = self.get_argument('draw', type=int)
-        start = self.get_argument('start', type=int)
-        length = self.get_argument('length', type=int)
-        search = self.get_argument('search[value]', type=str)
-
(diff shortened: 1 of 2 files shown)


flower/views/tasks.py
```diff
@@ -83,6 +83,40 @@ def key(item):
...
+            task_dict = as_dict(self.format_task(task)[1])
+            if task_dict.get('worker'):
+                task_dict['worker'] = task_dict['worker'].hostname
+
+            filtered_tasks.append(task_dict)
+
+        self.write(dict(draw=draw, data=filtered_tasks,
+                        recordsTotal=len(sorted_tasks),
+                        recordsFiltered=len(sorted_tasks)))
+
```
mher: Looks like `get` and `post` functions have identical body. It would be better to define `post` function with `post = get`.

johnarnold: Will do. thanks.

johnarnold: Done.

mher: Thanks @johnarnold . I will merge this PR after fixing the tests.

PR #158, opened 2013-12-04
Enhancement: Added feature to manually retry failed tasks. Added "Requeu...

jiffyjeff:
Enhancement: Added feature to manually retry failed tasks. Added "Requeue" button to task detail view; requeues the task with original args and kwargs using the Flower async-apply API. Supports an existing feature request (https://github.com/mher/flower/issues/17).

Issue #17, opened 2012-09-28
Feature request, retry failed task. 


jaredye:
It would be good if flower supports retrying failed task.

pedropregueiro:
👍

ryanprater:
+1

rafaelreuber:
+1

mher:
There are no plans to work on this feature at the moment until someone decides to contribute to the project.

PR #737, opened 2017-09-18
Enhancement: Added feature to manually retry failed tasks

tnir:
This PR is based on @jiffyjeff's work #158 with follow-up to be ready to be merged.

Enhancement: Added feature to manually retry failed tasks. Added "Requeue" button to task detail view; requeues the task with original args and kwargs using the Flower async-apply API. Supports an existing feature request (#17).

### DIFF
--- a/flower/static/js/flower.js
+++ b/flower/static/js/flower.js
@@ -325,6 +325,30 @@ var flower = (function () {
         });
     }
 
+    function on_task_requeue(event) {
+        event.preventDefault();
+        event.stopPropagation();
+
+        var taskname = $('#taskname').text();
+        var taskargs = $('#taskargs').text();
+        var taskkwargs = $('#taskkwargs').text();
+
+        $.ajax({
+            type: 'POST',
+            url: url_prefix() + '/api/task/async-apply/' + taskname,
+            dataType: 'json',
+            data: '{"args":' + taskargs + ', "kwargs":' + taskkwargs + '}',
+            success: function (data) {
+                var message = 'Requeued UUID: ' + data['task-id']
+                show_success_alert(message);
+            },
+            error: function (data) {
+                show_error_alert('Cannot enqueue task. In addition to broker, is Flower configured with correct ' +
+                    'Celery app/task config (e.g. --app or CELERY_APP)?');
+            }
+        });
+    }
+
     function on_task_terminate(event) {
         event.preventDefault();
         event.stopPropagation();
@@ -859,6 +883,7 @@ var flower = (function () {
         on_task_timeout: on_task_timeout,
         on_task_rate_limit: on_task_rate_limit,
         on_cancel_task_filter: on_cancel_task_filter,
+        on_task_requeue: on_task_requeue,
         on_task_revoke: on_task_revoke,
         on_task_terminate: on_task_terminate,
     };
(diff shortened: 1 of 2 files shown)


tnir:
@asfaltboy @mher @ask This is a follow-up against #158. I'd like you to review this PR.

tnir:
cc @embolalia @xrage @spudfkc

tnir:
@jiffyjeff What do you think of this PR?

SirEdvin:
Don't work, has problem with args parsing, because send tuple to json.

jheld:
@tnir @SirEdvin Would you like any help fixing/testing this?

tnir:
@jheld Yes if you can!

viiicky:
@tnir any update on this, if this is going to be merged?

clintonb:
What is preventing this from being merged?

tnir:
@clintonb As far as I know, someone does not think this PR works well.

jheld:
Last time I tried to use it, it didn't work. But I'm willing to try again.

jheld:
One extra issue is that args or kwargs repr could be truncated. If the worker has an increased repr size, that can be a work around -- but out of control of this PR.

Also, is there a way to remember/use any child tasks?

Additionally, if the task result backend is enabled and the args/kwargs are there (newer version of celery, not yet released), then we could take from that if it is available.

ryanprater:
+1

tnir:
Should I do something to get merged?

ryanprater:
@tnir yes please!

jheld:
If you do, please have a warning in docs or something that this do not re-queue child tasks, and depending on the size of args/kwargs, the re-queue itself might not run (repr may be truncated). The latter can be dealt with as part of running celery, but I don't think the former can.

jheld:
Another way to do this could be with use of the extended results feature on newer celery (4.3 I think), if all the settings are lined up. If the result backend has the tombstone and that extended setting is enabled, we should be able to get around the repr truncation issue at least.

mikolaje:
good feature, why can't it be merged?

jheld:
I think the idea is sound. But there are some caveats/documentation around it that I think would be required in this change should it be merged. If the truncation or extended results are not adjusted per the user, they may not be able to use the retry. That's effectively breakage from the user's perspective, unless there are prominently placed docs.

Due to how celery/events construct tasks and the flow in them, it makes this feature harder. Additionally, what happens if this task is part of a chain? We would like to ensure we can make it possible to call the rest of the chain should this task succeed on manual retry. That's my take, and I feel it's important to get it right if we can. If not, then yes, it could be left off for now.

jheld:
@mher would you be able to add me as a collaborator/maintainer? I'd like to help merge this (once we agree on readiness/impact).

mapineda:
Is there any update on this? @jheld

jheld:
I feel we're still blocked by at least the truncation issue. If the result backend extended setting support is merged then that will technically allow this to work better (so long as the key is still in the backend, not expired).

So although this is a useful change, it's sitting in a boundary where a lot of folks are going to have breakage, so long as the other pieces aren't addressed.

mher:
Closing this pull request because it will only work in limited cases

PR #739, opened 2017-09-21
Introduce TTLCache for the counter in the EventsState-class.

berkando:
This should invalidate worker that went offline automatically so that they are not show on the dashboard anymore.
Default value of the related parameter `cache_ttl` is chosen in a way that the behaviour of Flower should stay untouched.


mher:
`--purge_offline_workers` option already filters out offline workers

PR #766, opened 2018-01-03
Changed the url_prefix to rewrite the handlers regex patterns. This will

scrummyin:
cause nginx to no longer be needed to run in front of flower when adding
a url_prefix. A side effect will be that we no longer need to prefix
urls when we write them to the templates as the handlers will reflect
that change.


scrummyin:
Hey guys I have had this PR open for a bit and it looks to have generated a bit of interest. Can you let me know if there is anything I can do to improve the change if it's wanted? I undestand that it's not fully backwards compatible, but I was thinking it aligns more with the expected result. If you would like maybe I can make a second parameter with this new functionallity and keep url_prefix the same. Maybe a url_root parameter. Not sure how you would want the interaction there if both were set, but we can do something like that.

arthurpemberton:
@mher this change breaks existing Nginx installs. I don't see it documented anywhere either.

For anyone who runs into this, getting 404s on their Flower install after upgrading to 0.9.5 and using `--url-prefix` with Nginx, one no longer needs `rewrite ^/flower/(.*)$ /$1 break;` in your Nginx config.

scrummyin:
Hey @arthurpemberton, sorry about that, but I thought it was documented [here](https://github.com/mher/flower/blob/master/docs/config.rst#id30). With a straight upgrade with no changes to your flower setup should be good, but if you change your flower configuration when you upgrade, then nginx would need to be upgraded as well.

arthurpemberton:
@scrummyin I just did a straight `pip install -U` to flower, no config changes, other than restarting the service.

I didn't think to check every option that I use before upgrading in all fairness, I'm not sure breaking changes are best only documented in these kind of areas.

pembo13:
@scrummyin sorry, I changed to my primary profile from @arthurpemberton 

Before `--url-prefix` just affected URL presented to the browser. It didn't affect incoming URLs. so "/flower/dashboard/" needed to be rewritten to "/dashboard/" by Nginx. With the rewrite still in place, "/dashboard/" with `--url-prefix` gets 404ed.

joonhyungshin:
I was running multiple Flower instances with old version (0.9.3) behind a single Nginx reverse proxy, so now I cannot deploy more Flower instances with newer version (0.9.5). I like this fix, but too bad that the backward compatibility is broken in such a small [_patch_](https://en.wikipedia.org/wiki/Software_versioning#Degree_of_compatibility).

pembo13:
@joonhyungshin what's preventing you from doing multiple flower instances?

joonhyungshin:
@pembo13 When running behind the reverse proxy, the two versions have to be configured in a different way. I was using a single common proxy configuration for all Flower instances for the ease of deployment, but now I cannot do that unless I stick to 0.9.3 for all future deployments.

pembo13:
@joonhyungshin might want to share your confgs... that breakage seems like a lot.

joonhyungshin:
@pembo13 Sure, it looks something like this. (My actual config is different; just for demo purpose)
```
location ~* "/flower/([^/]+)/(.*)" {
    proxy_pass http://$1-flower.default.svc.cluster.local:5555/$2$is_args$args;
    proxy_set_header Host $host;
}
```
Yes, I'm running on k8s. The `default.svc.cluster.local` DNS is used internally by k8s, and this config allows me to connect to any Flower instance with Service `[NAME]-flower` via `http://URL/flower/[NAME]/`. Since 0.9.5, it shows 404.

pembo13:
@joonhyungshin this looks fixable... try passing the full URI to flower instead, and giving is the ful, appropriate `--url-prefix`, looks like you were giving it a partial `--url-prefix` before.

joonhyungshin:
@pembo13 Currently `--url_prefix` is set to `flower/[NAME]` for each instance. What should I change this to? Or is there any Nginx config that will work for both <=0.9.4 and >=0.9.5? Thanks for your help anyway.

pembo13:
@joonhyungshin you're not passing `[NAME]` ie. `$1` to the path, it should likely be something like `proxy_pass http://$1-flower.default.svc.cluster.local:5555/$1/$2$is_args$args;`

joonhyungshin:
That will probably work for >=0.9.5, but it won't work for old instances, which are <=0.9.4.

sudarshan2906:
Hi, I am doing the same which is as mentioned here - https://flower.readthedocs.io/en/latest/config.html?highlight=nginx#url-prefix
But, still getting `404: Not Found` in `<body>`
I am running `flower --url_prefix=flower` and my nginx config is 
```
server {
    listen 80;
    server_name example.com;

    location /flower/ {
        rewrite ^/flower/(.*)$ /$1 break;
        proxy_pass http://example.com:5555;
        proxy_set_header Host $host;
    }

}
```
The error comes in both when accessing `locahost:5555` and `localhost:80/flower`
flower version - `0.9.5`

pembo13:
@sudarshan2906 you should no longer need to `rewrite`

sudarshan2906:
@pembo13 Thanks. It worked. The documentation needs to be updated https://flower.readthedocs.io/en/latest/config.html?highlight=nginx#url-prefix

PR #772, opened 2018-01-17
Add task view support for result backend lookup on event cache miss

johnarnold:
This will allow looking up a task by id in the GUI even if the task is not in the event cache (due to flower restart or task cache size limit, etc)

E.g.
http://localhost:5555/task/07634992-5b5a-4228-b4c3-94bb184da2c5

Normally, you would just get a 'Unknown task' error.  This way you can get at least partial results for debugging.

### DIFF
--- a/flower/views/tasks.py
+++ b/flower/views/tasks.py
@@ -14,6 +14,10 @@ from tornado import web
 from ..views import BaseHandler
 from ..utils.tasks import iter_tasks, get_task_by_id, as_dict
 
+from celery.result import AsyncResult
+from celery.events.state import Task
+from celery.backends.base import DisabledBackend
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


johnarnold:
https://github.com/mher/flower/issues/542

jheld:
So as long as the return result of the task is currently still stored in the backend, then the idea is *something* should be able to be returned and shown by flower, in the event that flower no longer has its internal state for that task?

johnarnold:
@jheld yes that's the idea.  I've tested it, it works.  I'm still trying to get the "extended" task results into Celery, which will really make this useful.

sonirico:
What's the state of this? It would be a nice feature to fallback to a persistent, scalable result backend after flower restarts.

jheld:
This is an great addition. Can we please merge?

jheld:
@johnarnold I haven't recently reviewed the blockers/failing checks (could be minor stuff). Do you have time in the next month to take a peek?

mher:
Closing this pull request because it can cause many edge cases.

phillipuniverse:
@mher what are the edge cases?

Issue #784, opened 2018-03-06
Celery Flower doesn't work with Tornado 5.0.0


ksindi:
Calling [Flower API's](http://flower.readthedocs.io/en/latest/api.html#post--api-task-apply-(.+)) `/api/task/apply/*` is just hanging without the results returning, even though the task was processed and succeeded according to the Web UI.

Reproducing the error:
```
    # hangs even though task succeeded in logs
    curl -d '{"args": [1, 2]}' -H "Content-Type: application/json" -X POST http://0.0.0.0:5000/api/task/apply/tasks.add
```
Checking the task by id just shows it as pending even though the Flower UI shows it as a success and I also see the task result in the worker logs.
```
    curl -X GET http://0.0.0.0:5000/api/task/result/02fa2656-c92e-44e6-a2af-c04e666e3f89
    # {"task-id": "02fa2656-c92e-44e6-a2af-c04e666e3f89", "state": "PENDING"}
```
I suspect this log line is indicative of the issue: `celery-worker_1  | [2018-03-06 15:25:15,160: DEBUG/MainProcess] pidbox received method enable_events() [reply_to:None ticket:None]`.

There's also a `Traceback` in the flower side (see logs below) which happens when I hit Ctrl+C to get out of the hanging curl command above.

**tasks.py**
```
    from celery import Celery
    
    app = Celery('tasks', broker='redis://celery-redis:6379//0', backend='redis://celery-redis:6379//0')
    
    @app.task
    def add(a, b):
        return a + b
```

**Logs**
```
    celery-worker_1  | [2018-03-06 15:24:36,271: DEBUG/MainProcess] pidbox received method active() [reply_to:{'exchange': 'reply.celery.pidbox', 'routing_key': 'afe22ec9-5cde-330c-9306-c8dea457eecb'} ticket:48fb973b-51a1-4ad4-9482-04e31b19fb14]
    celery-worker_1  | [2018-03-06 15:24:36,282: DEBUG/MainProcess] pidbox received method reserved() [reply_to:{'exchange': 'reply.celery.pidbox', 'routing_key': 'd9911fb2-c681-3fe2-bdb2-248c14bf8351'} ticket:b69f2b6c-b7d7-4d26-a398-970ef22e4280]
[...]

PR #792, opened 2018-03-25
Fixes #791: use parse_version instead of tuple comparisons

### LINKED ISSUE #791: invalid literal for int() with base 10: '0rc1'
Hi, 
After a recent update to celery main branch: https://github.com/celery/celery/commit/c7f2a8b6111692696f69c682dd2c57010bae5b6a

This causes the following error:
```
File "/usr/local/lib/python3.6/site-packages/flower/events.py", line 70, in __init__
    if self.persistent and tuple(map(int, celery.__version__.split('.'))) < (3, 0, 15):
ValueError: invalid literal for int() with base 10: '0rc1'
```
This is the problematic line:https://github.com/mher/flower/blob/master/flower/events.py#L70

Which causes flower to crash completely.
maybe this can help for better parsing the version number:
https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python/11887885#11887885

pachewise:
Not sure how to add a test for this.

### DIFF
--- a/flower/events.py
+++ b/flower/events.py
@@ -11,6 +11,8 @@ from functools import partial
 
 import celery
 
+from pkg_resources import parse_version
+
 from tornado.ioloop import PeriodicCallback
 from tornado.ioloop import IOLoop
 
@@ -67,7 +69,7 @@ class Events(threading.Thread):
         self.enable_events = enable_events
         self.state = None
 
-        if self.persistent and tuple(map(int, celery.__version__.split('.'))) < (3, 0, 15):
+        if self.persistent and parse_version(celery.__version__) < parse_version("3.0.15"):
             logger.warning('Persistent mode is available with '
                            'Celery 3.0.15 and later')
             self.persistent = False


pachewise:
failures seem to be due to an existing issue (other PRs are failing in the same way, and tox run locally against master has the same issue)

barqshasbite:
Maybe due to Tornado 5? Issue #784 indicates that it is a problem. The project uses tornado>=4.2.0, so switching to tornado>=4.2.0,<5.0.0 may be required for now.

pachewise:
@barqshasbite thanks for pointing that out! Created PR #793 for fixing that issue

pachewise:
@auvipy do you know how we can reach @mher or another contributor with write access?

auvipy:
sorry no.

mher:
Merged, thanks! @pachewise please add your name to the list of contributors

david-caro:
Will there be a release for this fix?

PR #814, opened 2018-06-15
Added worker status filtering toolbar

louich:
With cloud native apps, worker are sometimes intended to have a short life and so the fact that we always see the dead workers as "offline" in the dashboard is kind of anoying. As I've seen it has been proposed (but refused)  to have the [ability to remove offline workers](https://github.com/mher/flower/issues/604) with the [PR 649](https://github.com/mher/flower/pull/649), which was determined to be too intrusive.

I propose a simple filter from radio buttons inserted into the datatable toolbar, which is blending with the current UI, easy to use and is of great use when we want to check either the "dead" or the "live" workers.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -77,3 +77,4 @@ Scott Kruger
 David Schneider
 S M Ahasanul Haque
 Leo Singer
+Louis-Charles Martin
--- a/flower/static/js/flower.js
+++ b/flower/static/js/flower.js
@@ -648,6 +648,7 @@ var flower = (function () {
             scrollX: true,
             scrollY: true,
             scrollCollapse: true,
+            dom: '<"toolbar">frtip',
             ajax: url_prefix() + '/dashboard?json=1',
             order: [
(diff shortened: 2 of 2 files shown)


dashdanw:
+1

williammcbroom:
This is a much needed feature for users in auto-scaling worker environments. It would be very helpful if this PR could be merged in.

davetuner:
+1

aambrozkiewicz:
👀

emoreth:
+1

rafaelreuber:
+1

jheld:
This seems like a really nice intermediate step/solution for cloud-native and the other code changes for the reset/clear of old offline workers.

They're both valid changes and this one is even less political!

@mher if the conflict is resolved, can we merge this?

jheld:
@mher If this conflict is resolved (or a PR based on it with conflicts resolved), can this be merged?

mher:
#852 introduces `--purge_offline_workers` option for hiding offline workers

PR #821, opened 2018-07-23
WIP: elasticsearch history support

jheld:
This looks to add elasticsearch as the task event history backend to flower.

# Proposed Solution
This is mainly 2 pieces:

- indexing events into elasticsearch in a somewhat efficient manner (for the process and for elasticsearch)
- searching/sorting/pagination of task history & task lookup by means of elasticsearch.

# Done so far (working)
- indexing tasks into elasticsearch

  - I have a background thread that buffers up task events based on a queue and sends bulk index requests into elasticsearch

- Searching (moderate support for different fields), sorting on all fields. The sorting appears to work but needs more QA. Pagination (works fairly well but I have noticed a bug in some circumstances).

- Dashboard able to pull from elasticsearch (at startup)

# Questions

- where will this logic live? flower subcommand? flower proper (w/ elasticsearch flag settings) @johnarnold 

 - originally I had the indexer outside of flower. It's now in flower, but configured in a hack-off standalone mode, based currently just on `argv`. There are a few `--elasticsearch` flags to control the behavior, just for draft/dev mode for now.
- I am using `kombu`'s `LRUCache` to cache certain `search_after` queries for the task history pagination. It is my way around the elasticsearch pagination restrictions, and keeps deep pagination requests super performant in my testing.

- can we improve the elasticsearch indexing process?

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -305,3 +305,48 @@ purge_offline_workers
 Time (in seconds) after which offline workers are automatically removed from dashboard.
 
 If omitted, offline workers remain on the dashboard.
+
+.. _elasticsearch:
+
+elasticsearch
+~~~~~~~~~~~~~
+
+Signals the process that it should use elasticsearch for history
+
+
+.. _elasticsearch_index_bulk_size:
+
+elasticsearch_index_bulk_size
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+How many documents will the elasticsearch indexer allow in a single bulk index API call.
+
+.. _elasticsearch_index_timeout:
+
+elasticsearch_index_timeout
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+How long should the background thread wait for the queue to fill up to elasticsearch_index_bulk_size
+
+.. _elasticsearch_day_retention:
+
+elasticsearch_day_retention
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For projects that require data retention management, this will specify how many days can have indexes at once.
+
+So if the value is 21, then any indexes older than 21 days will be deleted. This happens at startup and on day change.
+
+.. _elasticsearch_url:
+
+elasticsearch_url
+~~~~~~~~~~~~~~~~~
+
+Which URL is elasticsearch at?
+
+.. _elasticsearch_dashboard:
+
+elasticsearch_dashboard
+~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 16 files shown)


johnarnold:
@jheld will review as time allows.

johnarnold:
Improved search performance and capability is awesome. Elasticsearch is likewise good tech.  overall, I like where this is going.

Default behavior for the /api/task url is to use the celery backend to return task results.  The api is just a thin wrapper around the celery app -- same user experience basically as just using the python library with async result either Pending or returing the result. I would not switch this over to use Elasticsearch -- I would only use elasticsearch for the dashboard Task results view.   My reasoning is that there are a lot of reasons that the elasticsearch data could be missing or inaccurate -- missed events, ingestion delays, failure scenarios etc.  However, the celery backend should always have the tombstone record, or return Pending.

it would be great to refactor the whole event cache into a provider model.  Support the out-of-box in memory Event state, but also support other models, e.g. elasticsearch, mongodb, even in-mem but over a socket.  I have never liked the event state cache sharing fate with Flower -- it makes restarts painful, hard to keep it consistent across multiple flower instances, etc.

Another thing to think about is looking up a task by ID in the dashboard. If a task is not found in the flower "cache", flower task view can fall back to just getting the tombstone record from the backend.  See my PR https://github.com/mher/flower/pull/772

jheld:
Hey @johnarnold 

The reason that I've enabled elasticsearch for the search history, task UUID, and the search API, was so that we wouldn't be bound to tremendously slow queries/restart issues.

However, on my latest version of the PR, I have a way for the user to specify whether or not they want to pull from elasticsearch, programmatically. I hope this will satisfy the concern, at least conceptually, but maybe not?

Regarding the task by ID, I like your suggestion. That is, search from flower, then tombstone, and then from elasticsearch, as necessary. The problem is that I don't know how fast `get_task_by_id` is. If it's fast even with say 3 million (or more) records (in RAM), then it sounds like a no-brainer to hit it first, but if it takes 10 or more seconds, I'd rather pull from tombstone or elasticsearch if possible. Do you happen to know the performance characteristics?

As far a more pluggable set of alternative backends/framework for having a backend (including a socket-based RAM version), I love that idea.

I'm not sure if these would be installable extra pip-packages (extensions similarly to how `flask` or `bottle` work, I think), or if they would just come in as part of flower and can be enabled/disabled in a similar manner as to the way I've implemented it.

Please let me know your thoughts.

In terms of making it more pluggable, that will take extra work on this PR, which I'm totally willing to give my best shot at -- even taking new commits with it to improve it.

PR #829, opened 2018-08-08
WIP: Manual requeue on failed task

agungwk:
Add ability to requeue failed task


adimux:
In order for the task to be picked up by a worker, you have to specify a queue. It seems like we are not even specifying the queue here, so it will be requeued to the default queue I presume.

mher:
This will not work with long events and non-default queues

PR #834, opened 2018-08-23
Define dependencies using environment markers.

### LINKED ISSUE #833: Flower produces invalid wheels
Following on the heels of #674, flower has an additional issue where if a wheel is built on Python 2, it is invalid on Python 3, and vice-versa, due to the [conditionally-declared dependency on futures](https://github.com/mher/flower/blob/2f674aa4ed4f11243822869b70027c0b6f07a61c/setup.py#L49).

There are two possible approaches:

1. Remove the [universal wheel declaration](https://github.com/mher/flower/blob/2f674aa4ed4f11243822869b70027c0b6f07a61c/setup.cfg#L2), so that any wheels produced aren't considered compatible with both Python 2 and Python 3.

2. Use environment markers to declare the futures dependency on Python 2 only.

The former is less invasive, but the latter is probably preferable.

jaraco:
Fixes #833.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,3 +2,4 @@ celery>=3.1.0
 tornado>=4.2.0
 babel>=1.0
 pytz
+futures; python_version=="2.7"
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 import os
 import re
-import sys
 
 from setuptools import setup, find_packages
 
@@ -44,11 +43,6 @@ classes = """
 classifiers = [s.strip() for s in classes.split('\n') if s]
 
 
-install_requires = get_requirements('default.txt')
-if sys.version_info < (3, 0):
-    install_requires.append('futures')
-
-
 setup(
     name='flower',
     version=get_package_version(),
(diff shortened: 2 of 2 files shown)


jaraco:
Pretty sure the test failures are spurious.

eblume:
Hi, this just found its way downstream to the `poetry` python manager, here: https://github.com/sdispater/poetry/issues/752

Any word on when this might go live?

mher:
The test are still failing. Can you try to merge the master into the PR? It should fix the tests.

jaraco:
All it takes to re-merge a pull request with master is to close and re-open it (assuming there aren't conflicts).

jaraco:
Hmm. When I posted this PR, I was targeting it against the 0.9 branch. I don't remember why I did that - maybe because I was seeking to provide a fix that could be incorporated as an 0.9.x release... so merging to master was not the right thing to do... and tests are still failing on the 0.9 branch. I'm not sure what is the purpose of that 0.9 branch, but its CI builds are failing.

I see lots of commits to master but no new releases. I doesn't help much to target this at master if there's not going to be a new release of master.

So I've undone the merge with master.

This PR should be accepted as-is... and merged with master if appropriate. Or if 0.9 is abandoned, that branch should be removed so that bug fixes like this one can be targeted at master instead. As you can see from the output of the build, it is not failing at the point where these changes would have an effect.

We will continue to make internal private releases of v0.9.2+ with this patch.

Is there any point to the 0.9 branch? Is there a reason the project isn't making new releases with the contributions? Is there a reason 0.9 is ahead of master?

mher:
Thanks! I've merged #866. 0.9 branch is used for bug-fixes for 0.9.x versions.

PR #841, opened 2018-09-20
Document the GetQueueLengths API endpoint

asfaltboy:
It's pretty useful and I missed it until I started to dig into the code to copy the broker's tab into a new endpoint 🙄 

By the way it took me a while to figure out how docs work, would it be useful to add this to docs?:
```
pip install -r requirements/docs.txt
cd docs && make html
```


mher:
Thanks for the pull request @asfaltboy!  Yes, please add the build instructions to docs.

PR #842, opened 2018-09-25
Upgrade tornado

jmcarp:
Add a few fixes to make flower work with tornado 5.x:
* Use standard `run_in_executor` for background task in task apply
* Refactor auth providers to use coroutines instead of asynchronous decorator


jmcarp:
See https://github.com/tornadoweb/tornado/issues/2304#issuecomment-372152148 for details.

jmcarp:
Bump 😸

Issue #856, opened 2018-11-12
Refresh workers button not accessible


thenewguy:
I am observing "unknown worker" error when trying to view individual workers.  The worker list page works, but does not offer a refresh button.  If I pass ?refresh=True via the url, the page is reloaded and the links worked again.

This issue is related https://github.com/mher/flower/issues/546

CalebFenton:
Confirmed this bug is still happening with flower 0.9.3, celery 4.3.0, and using SSL for broker and result backend (not sure if that matters)

I couldn't get details on any of my workers -- they all said "unknown worker". The `?refresh=True` work-around fixed the problem, but it'd be nice if there was a button. I'm not sure if this is a simple bug (and I should PR a fix) or a new design direction.

PR #859, opened 2018-11-22
add back dashboard refresh button

adonisnafeh:
Fixes issue mentioned in #856. The refresh functionality is already there and is triggered by adding /?refresh=True to the url, I added a button that points to the current page with the refresh parameter appended to the url.

### DIFF
--- a/flower/templates/dashboard.html
+++ b/flower/templates/dashboard.html
@@ -17,6 +17,11 @@
     <div class="panel panel-default">
         <div class="panel-body">
 
+    <div class="btn-toolbar">
+      <div class="btn-group">
+        <a class="btn" href="./?refresh=True">Refresh</a>
+      </div>
+    </div>
     <table id="workers-table" class="table table-bordered table-striped">
       <thead>
         <tr>


jheld:
So this forces flower to re-setup its worker cache data, using the button? In this case, is your unknown worker actually online?

adonisnafeh:
The button reloads the page with the query string refresh=True which triggers [this code](https://github.com/mher/flower/blob/master/flower/views/dashboard.py#L31)  that updates the list of the workers in the app. The button was there previously but was removed in this [commit](https://github.com/mher/flower/commit/fe4277ad14162fd0b4a97c0ee3861ba0c7ca265f#diff-daf1a06897b23423690c155ca957bcd9 ). The commit removes the button from the interface but the code that performs the update is still there. 

The workers are online and you can see its activities in all the charts except for the Queued tasks. Clicking the button refreshes the list of workers and the activity starts showing up in that chart.

killthekitten:
Does not solve the #856 for my case

jheld:
Thanks for the merge. But yeah seems we still will have an issue with offline workers?

PR #864, opened 2018-12-18
Fixing GitHub OAuth callback handling

dinofizz:
There exists 2 issues:

- With the latest changes to auth.py `get_authenticated_user()` is being called with 3 parameters and no longer 4. This results in a TypeError complaining about a missing positional argument.

- The final `self.write()` line in `get_authenticated_user()` is not returning a Future that can be yielded by the calling coroutine. From Python 3.3 it is safe to use the "return" keyword but I opted for the Python 2.7 safe version and am returning a result by raising the "Return" exception (this is an accepted and valid way of returning values from coroutines in Tornado... which I learnt today)


ohld:
LGTM

PR #896, opened 2019-04-22
Fixed the issue of the broker page.

Ensteinjun:


flower/utils/broker.py
```diff
@@ -50,14 +50,16 @@ def __init__(self, broker_url, http_api, io_loop=None, **kwargs):
         self.io_loop = io_loop or ioloop.IOLoop.instance()
 
         self.host = self.host or 'localhost'
-        self.port = 15672
+        self.port = int(self.port or '15672')
```
mher: `urlparse` returns port as int, `self.port = self.port or 15672`

Ensteinjun: Thank you

Issue #886, opened 2019-03-07
Installing from PYPI and setup installs tornado 6.0.1


manmohana:
When installing flower==0.9.2 from PYPI its installing tornado-6.0.1.
I see in the code base we have the condition to install tornado version less than 6.0.0.
When trying to install from git repo code base it seems to be fine.
Thanks.

Kub-AT:
In requirements of release 0.9.2 there is "tornado>=4.2.0" 
https://github.com/mher/flower/blob/v0.9.2/requirements/default.txt

So you should add
tornado>=4.2.0,<6.0.0
in you project or install flower via git repo.

And maintainer of flower should release new flower with requirements fix but i don't see it coming.

mher:
A new flower version has been uploaded to pypi

Issue #878, opened 2019-02-19
Tornado no longer has asynchronous decorator


2stacks:
Looks like Tornado dropped this decorator in the latest release v6.0.0b1.

When trying to run 'celery flower' last message in stack trace is;

AttributeError: module 'tornado.web' has no attribute 'asynchronous'

jml:
AFAICT, this is fixed in code (see https://github.com/mher/flower/blob/master/requirements/default.txt), but that fix has not been released. In fact, there hasn't been a release for a year and a half (https://pypi.org/project/flower/#history), which makes me wonder whether I'm looking at the wrong page.

steverecio:
Is there a fix for this? I have flower pinned but this just took down my site

steverecio:
Had to add this to requirements.txt after flower:

`tornado>=4.2.0,<6.0.0`

2stacks:
I guess you could also install from main source branch but I also just
locked my requirements.

stevenmanton:
This dependency issue breaks stuff for me too. I wonder if there are any plans to release a new tagged version soon. Otherwise all the downstream projects will have to add hacks like @steverecio had to do with Astronomer.

mher:
I’m planning to release a new version next week

rmagee:
Seeing this as well.  Thanks.

shaohua:
10 hours spent on this issue as well

josiah14:
This issue broke our Airflow build, today, since we are using Celery.  Thanks, @2stacks for logging the issue, and @jml for identifying the cause.

I (very) strongly recommend adopting the practice of placing an upper bound on the versions of the dependencies your importing.  Otherwise, any bump in one of those dependencies to the next major version is highly likely to break your code (which is exactly what caused this issue).  As far as I can tell looking through the dependencies in the requirements directory, there's no upper-bound placed on any of the versions of the dependencies declared for this project.

astleychen:
@josiah14 totally agreed. Hopefully, the new release is coming soon so that I don't need to rollback to 0.9.1 as a workaround. Thanks.

josiah14:
@astleychen As a work-around, you can take advantage of the fact that Pip only allows one version of a library to be installed at a time.  So, you could `pip install flower` and then `pip install "tornado<6.0.0"`, as @steverecio pointed out in [this comment](https://github.com/mher/flower/issues/878#issuecomment-468964963).

josiah14:
[...]

PR #904, opened 2019-05-22
Added support for more tornado versions depending on python version

### LINKED ISSUE #903: Add support for tornado 6 with python 3.5.2 or higher
For python 3.5.2 or higher projects, need to add support for tornado >= 5.0.0 and < 7.0.0.
Currently,  after fixes for issues #878 and #886, `pip install flower` may actually downgrade tornado to 5.x which is not desirable.

I'm going to submit a PR to add environment markers and tox tests to support a better combination of tornado+python with flower.

I capped tornado at < 7.0.0 just to avoid a repeat of tornado breaking changes in the future without testing.

devmonkey22:
Added support for more tornado versions depending on python version:

- tornado>=4.2.0,<6.0.0; python_version<"3.5.2"
- tornado>=5.0.0,<7.0.0; python_version>="3.5.2"

This allows python 3.5.2+ projects to use the latest tornado (6.0.2 currently) with flower which was disallowed after fixes for issues #878 and #886.

Added tox tests to test combination of python, celery, and tornado versions.

This resolves #903.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -82,3 +82,4 @@ Bhargav Srinivasan
 Josiah Berkebile
 Deniz Dogan
 Aliaksei Urbanski
+Mike Dearman
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,6 +1,6 @@
 celery>=3.1.0; python_version<"3.7"
 celery>=4.3.0; python_version>="3.7"
-tornado>=4.2.0,<7.0.0; python_version<"3.5.2"
+tornado>=5.0.0,<6.0.0; python_version<"3.5.2"
 tornado>=5.0.0,<7.0.0; python_version>="3.5.2"
 humanize==0.5.1
 pytz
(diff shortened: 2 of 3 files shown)


mher:
@devmonkey22 thanks for the pull request! Are you using flower with tornado 6?

mher:
Versions in PR description and changes in default.txt don't match

devmonkey22:
Yes, I'm using flower 0.9.3 (with patch) on python 3.6 with tornado 6.0.2 without known issues now.
Sorry about the version mismatch - I accidentally left my test to prove python 2.7 would error with tornado 6. Thanks for catching that - would defeat the purpose of the whole PR :)

vijaykramesh:
I'm a bit confused about this change, this breaks the fix in https://github.com/mher/flower/issues/784 and as far as I can tell, oauth won't work anymore...  @devmonkey22 have you tried using an --auth_provider with this? I get `RuntimeError: Cannot run the event loop while another loop is running`

devmonkey22:
@vijaykramesh I don't have an oauth provider to test with. This change only allows more tornado versions to be installed in a project. If you need to constrain your environment to a certain tornado version within this range, I'd recommend to use a constraints file, explicit version specifier in your requirements file, etc.

PR #916, opened 2019-07-17
Default bind on localhost

gthess:
The documentation hints that flower binds to localhost by default.
Currently the address option defaults to the empty string. This option is passed to tornado which in turn will handle it as a signal to bind to all interfaces.
From https://github.com/tornadoweb/tornado/blob/master/tornado/netutil.py
>def bind_sockets(
>...
>Address may be an empty string or None to listen on all available interfaces...

This PR addresses the issue by specifically using 'localhost' as the default value.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -83,3 +83,4 @@ Josiah Berkebile
 Deniz Dogan
 Aliaksei Urbanski
 Mike Dearman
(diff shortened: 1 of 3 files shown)


mher:
Binding to all interfaces is desired default behavior. Localhost appears only in startup log

PR #917, opened 2019-07-19
Improving the view kwargs and args in task

rodrigopmatias:
I would like you to evaluate this PR (# 415) since in 2015 this same code had already been accepted sooner afterwards for one mistake it was overwritten by another PR.

This modification makes the viewing of args and kwargs in the visualization of tasks more humane.


rodrigopmatias:
@mher What was the problem with PR, is there any way to make its adoption possible?

Issue #787, opened 2018-03-14
Flower as CentOS service with is_persistent flag and db configurations doesn't load dashboard data and can't open worker infor


farhan1988:
Hi Team,

I have created a flower service in my local 

**/etc/systemd/system/flower.service**
```
[Unit]
Description=Flower Celery Service

[Service]
User=vagrant
Group=vagrant
WorkingDirectory=/home/vagrant/
ExecStart=/usr/bin/flower --/var/log/supervisord.log
Restart=on-failure
Type=simple

[Install]
WantedBy=multi-user.target
```

**/home/vagrant/flowerconfig.py**

```
port		=	5000
broker 		=	'redis://:<password>@192.168.100.20:6379/0'
broker_api 	=	'redis://:<password>@192.168.100.20:6379/0'
basic_auth 	=	['sun:flower']
persistent 	=	True
db		=	'/home/vagrant/flower.db'
```

status of service is **Green**
```
[vagrant@node1 ~]$ sudo systemctl status flower
● flower.service - Flower Celery Service
   Loaded: loaded (/etc/systemd/system/flower.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2018-03-14 10:43:15 UTC; 7min ago
 Main PID: 6578 (flower)
   CGroup: /system.slice/flower.service
           └─6578 /usr/bin/python2 /usr/bin/flower --/var/log/supervisord.log

Mar 14 10:43:18 node1 flower[6578]: [u'celery.accumulate',
Mar 14 10:43:18 node1 flower[6578]: u'celery.backend_cleanup',
Mar 14 10:43:18 node1 flower[6578]: u'celery.chain',
Mar 14 10:43:18 node1 flower[6578]: u'celery.chord',
Mar 14 10:43:18 node1 flower[6578]: u'celery.chord_unlock',
Mar 14 10:43:18 node1 flower[6578]: u'celery.chunks',
Mar 14 10:43:18 node1 flower[6578]: u'celery.group',
Mar 14 10:43:18 node1 flower[6578]: u'celery.map',
Mar 14 10:43:18 node1 flower[6578]: u'celery.starmap']
Mar 14 10:43:18 node1 flower[6578]: [I 180314 10:43:18 mixins:224] Connected to redis://:**@192.168.100.20:6379/0
```

Everything so far works fine.. Except for two issues

Issue #1 -->  Unable to see Dashboard data after the service refresh but **/tasks** shows list of data from previous executions..

**Dashboard**
[...]

PR #926, opened 2019-08-24
Add ability to dump and restore state counter from persistent database (shelve)

sshishov:
When use `persistent=True` to dump state to the file, counter data is not dumped.

Therefore after restoring from the file, on dashboard we do not see any statistics, everything is 0.

To add ability to store counter along with the state, `__reduce__` method was overridden and `__init__` to accept counter.

Note: we can try to restore from `kwargs` but for these we need to create `__new__` constructor for EventState class and add additional magic method (as shelve is using protocol 3 by default)

### DIFF
--- a/flower/events.py
+++ b/flower/events.py
@@ -34,8 +34,9 @@ class EventsState(State):
     # EventsState object is created and accessed only from ioloop thread
 
     def __init__(self, *args, **kwargs):
+        args, extra_args = args[:10], args[10:]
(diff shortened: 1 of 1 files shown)


sshishov:
Fixes #787 partially (dashboard issue)

sshishov:
Very soon it will be anniverasy of 5 years 🎉

mher:

Issue #928, opened 2019-09-06
Incorrect messages count displayed on Broker page when using priority queues with Redis


sgallen:
In the class `Redis`, `super` is called without passing `*args` and `**kwargs`, as a result `'priority_steps'` is not passed to the base and the `DEFAULT_PRIORITY_STEPS` are always used.  If messages are queued using priorities 0, 3, 6, or 9 the count is correct but if you configure the priorities to be, for example, 0 through 9, messages not in the default won't contribute to the count.

PR #929, opened 2019-09-06
Fix incorrect message count when using priority queues (#928)

sgallen:


mher:
Thanks for the pull request! Can you add a test?

sgallen:
I can definitely add a test. Would you suggest an extension to an existing test in test_broker.py or a new separate test in there?

mher:
I'd add a new `test_priority_steps ` method to https://github.com/mher/flower/blob/master/tests/unit/utils/test_broker.py#L46.

PR #930, opened 2019-09-13
Adding setting option to ignore a list of queues

cdknorow:


mher:
This behavior can be implemented with worker/broker routing configuration

PR #953, opened 2020-01-03
Feature: view for queued tasks on broker

jheld:
This will help solve some places where either the team doesn't want to use the celery CLI (inspect or events), or doesn't have access to it. This specifically looks to have a simple but usable interface to seeing what tasks are queued on the broker, with pagination, for a given queue, and nothing more.

Being able to filter by task and time etc would be prohibitive considered that at least redis doesn't have a native way for us to do that, using a list. Using redis streams or sorted sets, could alleviate portions of that, but that is _way_ outside of scope since those are not currently used in the redis broker implementation at all.

It would be useful to have feedback on what sort of information (or even additional views/pages) we should show on this page. We can add more to the view and I've added the colVis DataTable extension to help with the dynamism of showing columns on the fly. `body` for instance is great so we can decode the payload, but normally we likely don't want to always see it because it may be large to show on the page.

RabbitMQ is currently not supported (partly due to the documentation not being easy for me to follow, but also I don't have that broker installed though I could still test it a little).


mher:
Closing this pull requests because this feature will not work with RabbitMQ

PR #955, opened 2020-01-27
Add support for RedisSentinel broker

fjcapdevila:
Add basic support for Redis + Sentinel as broker.

### DIFF
--- a/flower/utils/broker.py
+++ b/flower/utils/broker.py
@@ -200,8 +200,9 @@ class RedisSentinel(RedisBase):
         return master_name
 
     def _get_redis_client(self):
+        connection_kwargs = {'password': self.password}
         # TODO: get all sentinel hosts from Celery App config and use them to initialize Sentinel
-        sentinel = redis.sentinel.Sentinel([(self.host, self.port)])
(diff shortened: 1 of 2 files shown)


flower/utils/broker.py
```diff
@@ -203,6 +246,8 @@ def __new__(cls, broker_url, *args, **kwargs):
             return RedisSsl(broker_url, *args, **kwargs)
         elif scheme == 'redis+socket':
             return RedisSocket(broker_url, *args, **kwargs)
+        elif scheme == 'sentinel':
```
mher: How about using `redis+sentinel` as a protocol name?

fjcapdevila: I've decided to use that protocol name for consistency with Celery docs:
https://docs.celeryproject.org/en/latest/getting-started/brokers/redis.html#configuration

doncraig:
Is there any documentation available on how to configure master_name in Docker?

vuducmanh11:
> Is there any documentation available on how to configure master_name in Docker?

are you configured it? :(

Issue #960, opened 2020-02-29
python3 only support


jheld:
Although celery 5 is where py2.7 support will be removed, does flower have the same burden?

syntax wise so long as we don't use anything special that celery internally cannot, it should be compatible? and since we have a test suite perhaps that will keep us honest?

i'm going to open a PR but not necessarily expecting it will get merged short term.

mher:
I'm closing this issue since there is an open pull request for it

PR #961, opened 2020-02-29
Python 3.5-3.8 support; celery 4.4 support; adjust CI.

jheld:
this looks to support python3.5-3.8 (and include celery 4.4 and tornado 6.0 in the tests).

this is not looking to change code patterns/styles for py3 era only. This is for issue #960 

it's possible that we're not yet ready to make this commitment, but i suggest that any discussion around timing (not the code change itself) be pushed down to the issue level. celery has to support its community (wider than flower), but I don't think (I may be wrong) that flower has to be caught up in that version issue as strongly. we just must make sure that the tests are honest. and for another, why burden ourselves with the possibility of any security issues from py2.7?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,9 +8,20 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7]
-        celery-version: [4.1, 4.2, 4.3]
-        tornado-version: [5.0, 5.1]
+        python-version: [3.5, 3.6, 3.7, 3.8]
+        celery-version: [4.1, 4.2, 4.3, 4.4]
+        tornado-version: [5.0, 5.1, 6.0]
+        exclude:
+          - python-version: 3.7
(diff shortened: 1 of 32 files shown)


jheld:
@mher thoughts?

karolpawlowski:
Is anyone gonna to release this?

> ERROR: flower 0.9.3 has requirement tornado<6.0.0,>=4.2.0, but you'll have tornado 6.0.4 which is incompatible.

I want to update tornado within my project. I have python 3.7.7

jheld:
I don't think I was necessarily attempting to keep tornado under 6.0.4. I suppose we can try to update that pin.

I'd prefer the tornado upgrade/less strict occur before this PR, but we can include it here first for testing.

karolpawlowski:
Ok, I see that tornado>6.0 is already implemented and merged but pip contains lib version from a year ago

mher:
Looks good. Thanks for the pull request @jheld . I will merge it for the next release

PR #965, opened 2020-03-13
Add Okta Login Handler

logston:
The goal of this PR is to allow for the use of Okta as an OAuth 2.0 provider.


mher:
Thanks for the pull request. Please add unit tests

PR #968, opened 2020-04-10
use brighter red and green for important and success labels

jadami10:
Our team discovered the existing red/green colors are too muted for those with red/green color blindness to discern. This diff updates those colors to be fairly brighter.
image
image


mher:
The https://github.com/mher/flower/blob/master/flower/static/css/flower.css should be used for css overrides. Otherwise the changes will get lost with the next bootstrap upgrade.

PR #969, opened 2020-04-21
fix google auth

sokovnich:
It fixes a problem with google auth.
Login failure message "Invalid parameter value for approval_prompt: 'auto' is not valid".

Similar issue
https://github.com/thephpleague/oauth2-google/issues/87


mher:
Please provide a description

sokovnich:
> Please provide a description

done

PR #979, opened 2020-05-26
Remove explicit humanize version

jmsmkn:
Humanize is now version 2.4.0, the only major change is dropping Python 2 support https://github.com/jmoiron/humanize/releases

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,5 +2,5 @@ celery>=3.1.0; python_version<"3.7"
 celery>=4.3.0; python_version>="3.7"
 tornado>=5.0.0,<6.0.0; python_version<"3.5.2"
 tornado>=5.0.0,<7.0.0; python_version>="3.5.2"
-humanize==0.5.1
+humanize
 pytz


mher:
What issue this change is supposed to solve?

jmsmkn:
I cannot install flower into the environment that I want to run it in as that depends on an up to date version of humanize.

hughsie:
I'm also in the same boat; I have to install flower manually, then lvfs-website using requirements.txt to get an up-to-date version of humanize.

mher:
Please update the version then. We use explicit version numbers to avoid incompatibility problems in future dependency releases.

jmsmkn:
I don't see the need, you would catch this in CI, pinning this makes it more difficult for the rest of us. `pytz` is not pinned either and the tests cover the part of the public API that are used in this application from Humanize (which is only one call to `naturaltime` anyway): https://github.com/mher/flower/blob/ceb6e1bd8c432f7783ef6345485d1abd6c7f492b/tests/unit/utils/test_template.py#L6

PR #981, opened 2020-06-04
Fix redis SSL broker handling

mbierma:
When the `BrokerView` instantiates a [Broker instance](https://github.com/mher/flower/blob/master/flower/views/broker.py#L32), the URI is passed in using kombu's `as_uri` method. `as_uri` will convert a redis connection string from `rediss://localhost:6379/0` to `redis://localhost:6379/0` (removing the second 's'). This results in the scheme check to never hit the `scheme == 'rediss'` case. If a redis SSL broker is used, this results in hanging requests to the /broker webpage.

This update should fix the issue by using the `broker_use_ssl` kwarg to determine if the redis broker is using ssl, in addition to the `rediss` scheme.

### DIFF
--- a/flower/utils/broker.py
+++ b/flower/utils/broker.py
@@ -136,6 +136,7 @@ class RedisBase(BrokerBase):
 class Redis(RedisBase):
 
     def __init__(self, broker_url, *args, **kwargs):
+        self.ssl = 'broker_use_ssl' in kwargs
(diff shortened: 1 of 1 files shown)


mher:
`as_uri` doesn't modify protocol

```
>>> from celery import Celery
>>> Celery("tasks", broker="rediss://").connection().as_uri()
'rediss://localhost:6379//'
>>> celery.__version__
'4.4.3'
```

PR #433, opened 2015-07-18
enabling cors by adding tornado cors mixin

asmodehn:
simple change, just following tornado-cors doc made it work for me...
This is a cherry pick from a change that I tested on top of 0.8

Let me know if I should make more changes...

Issue #429, opened 2015-07-18
Flower API and CORS


asmodehn:
I am trying to use the flower API as a REST backend to a webapp frontend ( potentially sitting on a different host ) that I am writing now.
However I am having issues with CORS requests. It seems to me that the preflight request is not handled properly : 

```
[W 150718 11:57:01 web:1811] 405 OPTIONS /api/tasks (127.0.0.1) 19215.00ms
```
```
[E 150718 11:51:49 web:931] Uncaught exception in write_error
    Traceback (most recent call last):
      File "/opt/ros/indigo/lib/python2.7/dist-packages/tornado/web.py", line 929, in send_error
        self.write_error(status_code, **kwargs)
      File "/home/alexv/ROS/gopher_concert/src/flask-ext-catkin/src/flower/views/__init__.py", line 48, in write_error
        self.write(message)
      File "/opt/ros/indigo/lib/python2.7/dist-packages/tornado/web.py", line 656, in write
        raise TypeError("write() only accepts bytes, unicode, and dict objects")
    TypeError: write() only accepts bytes, unicode, and dict objects
[W 150718 11:51:49 web:1811] 405 OPTIONS /api/task/async-apply (127.0.0.1) 2579.07ms
```

So, should I be able to make CORS requests to flower backend ?
Or is there something I am missing here ? 
By the way, I'm using v0.8.2

asmodehn:
Also after I managed to disable the preflight request by using "text/plain" as the contentType of my request, I still cannot process the reply as : 
`XMLHttpRequest cannot load http://localhost:5555/api/tasks. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8888' is therefore not allowed access.`

asmodehn:
I added tornado-cors mixin to BaseHandler and it seems to have done the trick...
Please let me know if this is acceptable for you and I ll do a pull request for it once I have tested it more.

asmodehn:
https://github.com/mher/flower/pull/433

mher:
Closing this issue. There is a pull request #433

tiktuk:
I am having the same issue, could be nice to have the pull request pulled :) .

nevelis:
Guess we should've closed the issue after the pull request was merged! 👍

PR #994, opened 2020-06-27
Added CORS support

davidbunk:
Adding support for CORS, an issue described in #429 , and solved for older versions in #433 .
The module tornado-cors in #433 is outdated and does not support tornado>=5.11, this solution does not depend on additional dependencies by only adding the needed headers for CORS to BaseHandler.

Let me know if there's any more required changes on this!

### DIFF
--- a/flower/views/__init__.py
+++ b/flower/views/__init__.py
@@ -17,6 +17,15 @@ logger = logging.getLogger(__name__)
 
 
 class BaseHandler(tornado.web.RequestHandler):
+    def set_default_headers(self):
+        self.set_header("Access-Control-Allow-Origin", "*")
+        self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+        self.set_header('Access-Control-Allow-Methods', ' PUT, DELETE, OPTIONS')
+
+    def options(self):
+        self.set_status(204)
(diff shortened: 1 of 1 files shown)


Charles-YYH:
Hello, I've got a question regarding line 25, see comments below
```python
def options(self): # Is there any reason to not add *kwargs here ?
  self.set_status(204)
  self.finish()
```
Pointing out because when sending OPTIONS request to an url like http://example.url/api/task/async-apply/task_name

There will be an error
> HTTPServerRequest(protocol='http', host='***', method='OPTIONS', uri='/api/task/async-apply/****', version='HTTP/1.1', remote_ip='172.22.0.1')
      File "/home/vas/.local/lib/python3.8/site-packages/tornado/web.py", line 1702, in _execute
        result = method(*self.path_args, **self.path_kwargs)
    TypeError: options() takes 1 positional argument but 2 were given

Are we deliberately forbidding this kind of requests or is there any reason not doing this?
Thanks

Charles-YYH:
Hello, just curious. Is there any reason of not including POST in the field 
`Access-Control-Allow-Methods` ? Sometimes it might be helpful if one can start a task from frontend. 
Thanks.

Issue #945, opened 2019-12-02
FEATURE: Add health check for web server


scarbeau:
## Use case
Running Flower behind ALB within AWS.

## Implementation

/flower/views/alive.py
```python
from __future__ import absolute_import

from ..views import BaseHandler


class HealthCheckHandler(BaseHandler):
    def get(self):
        self.set_status(200)
        self.write('OK')
```

/flower/urls.py
```python
(r"/alive", HealthCheckHandler),
```

/tests/views/test_alive.py
```python
from tests.unit import AsyncHTTPTestCase


class AliveTests(AsyncHTTPTestCase):
    def test_health_check(self):
        r = self.get('/alive')
        self.assertEqual(200, r.code)
```

jordanmkoncz:
This is a feature that we currently need too. We're deploying a Django backend with Celery/Flower to AWS; we're using Fargate and ELB and as a result we need Flower to provide a healthcheck using HTTP on port 80.

engin-bulut:
hey @jordanmkoncz We had the same problem. But, as you know all we need for the health check is one endpoint returns 200 response. We use `/api/workers` endpoint of flower as a health check path and it worked. Give it a try.

mher:
You can use `/metrics` endpoint for health checks

jordanmkoncz:
@mher can you please clarify? When I call `/metrics` or `/api/metrics` I get a 404, and https://github.com/mher/flower/blob/master/flower/static/swagger.json appears to have no mention of a "metrics" endpoint.

mher:
`/metrics` endpoint https://github.com/mher/flower/blob/master/flower/urls.py#L79

jordanmkoncz:
@mher thanks for the clarification. Looks like this has just been added recently to `master` and has not yet been included in any release. I assume this will be part of an upcoming `0.9.5` release, so I'll try it out then.

pblankley:
I know this issue is closed, but for those of you who still have this problem (e.g. anyone who uses auth), you can use `/healthcheck` and that should work swimmingly for you :)

OmarElsebaey001:
@pblankley you saved my day with this comment , using auth here , thanks a million

PR #1012, opened 2020-07-25
Add healthcheck endpoint

bmwant:
### Rationale

When running Flower with auth enabled using some orchestration service it's useful to have `healthcheck` endpoint which indicates status of the application without authN required.

### Examples
a)
AWS healthcheck for target group

Screen Shot 2020-07-25 at 7 16 20 PM
Screen Shot 2020-07-25 at 7 17 37 PM

b)
[docker-compose healthcheck](https://docs.docker.com/compose/compose-file/#healthcheck)
image
Screen Shot 2020-07-25 at 7 22 42 PM

### DIFF
--- a/flower/urls.py
+++ b/flower/urls.py
@@ -68,6 +68,7 @@ handlers = [
     (r"/api/task/events/task-custom/(.*)", events.TaskCustom),
     # Metrics
     (r"/metrics", monitor.Metrics),
+    (r"/healthcheck", monitor.Healthcheck),
     # Static
     (r"/static/(.*)", StaticFileHandler,
      {"path": settings['static_path']}),
--- a/flower/views/monitor.py
+++ b/flower/views/monitor.py
@@ -11,3 +11,9 @@ class Metrics(BaseHandler):
     @gen.coroutine
     def get(self):
         self.write(prometheus_client.generate_latest())
+
+
+class Healthcheck(BaseHandler):
+    @gen.coroutine
+    def get(self):
+        self.write("OK")
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,5 +1,6 @@
 celery>=3.1.0; python_version<"3.7"
 celery>=4.3.0; python_version>="3.7"
+vine==1.3.0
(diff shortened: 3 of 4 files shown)


bmwant:
@mher let me know if that's something that can be added, so I can update unittests and docker-compose file [like here](https://github.com/tobybatch/kimai2/blob/main/docker-compose.yml#L32)

mher:
`/metrics` endpoint can be used for the health check

mher:
Duplicate of https://github.com/mher/flower/issues/945

bmwant:
@mher the main difference is that `metrics` require `@web.authenticated`, so basically I need to send Auth header with credentials included for the healthcheck request which is not always possible. Is there a way to deal with that if I'm running flower with `celery flower -A app --basic_auth="user:pass"`?

MarsssBlkm:
Having the same problem running Airflow with celery executor and flower. I have to set health check response as 401 as well for now.

lggwettmann:
Same here, would be great if there would be an auth-free endpoint. Which alternatives are people using now?

imressed:
@mher could you please share some more details regarding the health check and `/metrics` endpoint? 
As I can see metrics endpoint requires auth headers to be provided in order to get a response (and not 401 response) (i mean it has header `@web.authenticated`) 
Can it be a possible option to extend health-check functionality by adding an auth-free endpoint like proposed in this PR? 
I've already seen a couple of the same requests. Thanx for the update on this!

mher:
Please add unit tests for the `/healthcheck` endpoint and I will merge the PR.

imressed (approved):

lggwettmann:
Awesome guys and thanks for the fast work. Any plan on when you'll release this? :)

bmwant:
@lggwettmann @mher added unittest, should be ready to go

bmwant:
It is failing due to https://github.com/mozilla/MozDef/issues/1678. I can downgrade the lib but let's ask @mher about the proper action to do here

mher:
@bmwant feel free to downgrade the version

Issue #891, opened 2019-03-10
Celery's priority seperator has become configurable


jobec:
It's now possible to configure the seperator when using Redis as a broker. See https://github.com/celery/kombu/pull/1020

This means that hard coding this separator might break things:
https://github.com/mher/flower/blob/2da8d7dbabffdb5c6fd9aa141946e1881bb9281c/flower/utils/broker.py#L101
https://github.com/mher/flower/blob/2da8d7dbabffdb5c6fd9aa141946e1881bb9281c/flower/utils/broker.py#L120

mher:
Feel free to open a pull request

dejlek:
Why? So that we wait for half a year for it to be reviewed and merged?

Asday:
> Why? So that we wait for half a year for it to be reviewed and merged?

...  Because this is how open source works?  The author of a package shouldn't have to be expected to do all the work.

mher:
Fixed in ecc18106cb17fc62f592473a73c30a2747a1e749

PR #1014, opened 2020-08-01
FLOWER-891 respect custom separator

meeech:
#891 Respect the configured separator broker transport option for redis

I've marked this as a draft because I would like some feedback. I'm still testing it out locally (I'm having some issues, but not sure if its flower, or something else I've done wrong with custom separator on my celery workers

I'm seeing 
```
[W 200731 23:30:17 inspector:42] Inspect method reserved failed
[W 200731 23:30:17 inspector:42] Inspect method revoked failed
[W 200731 23:30:17 inspector:42] Inspect method stats failed
[W 200731 23:30:17 inspector:42] Inspect method conf failed
[W 200731 23:30:17 inspector:42] Inspect method registered failed
[W 200731 23:30:17 inspector:42] Inspect method scheduled failed
[W 200731 23:30:17 inspector:42] Inspect method active failed
[W 200731 23:30:17 inspector:42] Inspect method active_queues failed
```
but this is even with master flower without my changes. 
My worker shows this error `[2020-08-01 03:30:16,361: ERROR/MainProcess] Control command error: ValueError('too many values to unpack (expected 3)',)` (not your issue - just giving some context :D)

### DIFF
--- a/flower/utils/broker.py
+++ b/flower/utils/broker.py
@@ -94,7 +94,7 @@ class RabbitMQ(BrokerBase):
 
 
 class RedisBase(BrokerBase):
-    SEP = '\x06\x16'
+    DEFAULT_SEP = '\x06\x16'
     DEFAULT_PRIORITY_STEPS = [0, 3, 6, 9]
 
     def __init__(self, broker_url, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


mher:
The separator change looks good

meeech:
@mher So, I'm comfortable with this PR now. I've tested it out and it works fine with custom seps like `-x-x-` and `:`.

But I could use some guidance on one issue - maybe you have some insight or tip what I should look at (you know the code way better than me! :D ) I'm not even sure if this is a celery, kombu, redis or flower issue (though most likely a combo of all these, and how I'm using it?)

The issue I found is when using a `.` as the custom sep, then things start to break with regards to flower. My queue's seem fine, and everything is working. But I can't monitor using flower when the sep is a `.` I see the worker flapping online/offline, can't click through to the worker (gives 404). The `broker` tab doesn't even show all my queues (i'm running 2 queues).

I get those messages about the `Inspect method`s failing, but not clear to me why they fail, or whats happening under the hood there. My suspicion is something to do with how redis works with KEYS which accepts regex patterns - so maybe the `.` would have to be escaped? (again, only a guess here). But otherwise, my celery queues themselves seems to be working fine, processing the jobs. 

In any case, I would be fine just putting a warning/caveat about not using `.` as your custom sep else flower can have issues monitoring. 

Some other details that may be relevant - I'm just using celery worker command to consume tasks from redis broker `celery worker -O fair -Q jobs`

Anyhow, let me know how you would like to proceed.

meeech:
@mher not sure if you get notified when I edit a comment, but the issue I alluded to above is resolved, so this should be gtg.

LostVector:
Can we get clarity how this is supposed to actually be set?  Is it possible to customize the separator and priority steps from the command line or environment?

PR #1028, opened 2020-09-22
Add periodic state save.

### LINKED ISSUE #803: Periodic state save to DB
Currently the only time that the state is written to DB is on SIGTERM.

In a docker setup it can be pretty difficult to ensure that the signal is always received on shutdown. This also means that if something goes wrong, the state is lost.

It would be really useful if instead flower would at least periodically write to db.

AntonP21:
Resolve #803

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -220,13 +220,13 @@ port
 
 Run the http server on a given port (by default, `port=5555`)
 
-.. _state_save_timeout:
+.. _state_save_interval:
 
-state_save_timeout
+state_save_interval
 ~~~~~~~~~~~~~~~~~~
 
-Sets the timeout for saving state. state_save_timeout=0 means
-that periodic saving is disabled (by default, `state_save_timeout=0`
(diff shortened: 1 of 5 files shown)


ghedb (commented):
Im not super familiar with Tornado but everything looks good to me from a logical perspective.

I just have one comment, does the word timeout really make sense? 

Maybe state save interval would be more clear?

ghedb:
> Im not super familiar with Tornado but everything looks good to me from a logical perspective.
> 
> I just have one comment, does the word timeout really make sense?
> 
> Maybe state save interval would be more clear?

I just realized that it _is_ a timeout, its the tornado argument that I find a bit unclear.

AntonP21:
> Im not super familiar with Tornado but everything looks good to me from a logical perspective.
> 
> I just have one comment, does the word timeout really make sense?
> 
> Maybe state save interval would be more clear?

I agree that the word "interval" is clearer than "timeout"

PR #1048, opened 2020-12-11
Proposal: use jinja for flower templates in preparation for adding flask app support

avilaton:
I am trying to get flower running as a flask blueprint and I realized we could change templates to jinja so they can be shared. I'm submitting this PR both for your consideration and to check if it breaks CI.
I also would like to emphasize that celery is migrating commands to `click`, so other pallets dependencies might be coming into flower, which is why I think this might be acceptable.


mher:
Jinja support is not currently planned

Issue #1053, opened 2020-12-24
Vine pinned == 1.3.0 in 0.9.7 is it necessary since it breaks dependencies


campbellmc:
**Describe the bug**
It seems vine was pinned to 1.3.0 in requirements/defaults.txt in this PR; "Add healthcheck endpoint (#1012)"

There is no real explanation to why it was pinned - there was an issue with vine at the time with vine/amqp which affected celery.

Seems like this is dependency that should not really be pinned here.

**To Reproduce**
Start a project with the following requirements.txt:
celery[redis]==5.0.5   
flower==0.9.7
amqp==5.0.2
vine==5.0.0 

This won't resolve dependencies for installation.
Pinning flower to 0.9.5 will allow the build and everything works fine.

**Expected behavior**
Should be able to resolve dependencies and build image with vine==5.0.0

**Screenshots**
n/a

**System information**
n/a

mher:
Thanks for opening the issue @campbellmc. Feel free to open a pull request to remove it.

hakamadare:
@mher it looks like #1056 that fixes this issue has been merged, do you have an idea of when you'll publish an updated package that contains this fix?

bmyrtil-fsns:
I dont believe that merge actually fixed it. the vine dependency is pinned twice for some reason but only one of the lines were removed... unless im looking at the file wrong.

hackslashX:
> I dont believe that merge actually fixed it. the vine dependency is pinned twice for some reason but only one of the lines were removed... unless im looking at the file wrong.

You're right. It's mentioned there twice. Both entries have to be removed.

mher:
Fixed in the master

PR #1056, opened 2021-01-02
remove vine==1.3.0 to solve depndency break with celery==5.0.5

polarAli:
Fix issue #1053 by removing vine==1.3.0 from requirements/defaults.txt

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,7 +1,6 @@
 celery>=4.3.0,<5.0.0; python_version>="3.7"
 vine==1.3.0
 tornado>=5.0.0,<7.0.0; python_version>="3.5.2"
-vine==1.3.0
 prometheus_client==0.8.0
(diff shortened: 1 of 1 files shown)


bmyrtil-fsns:
Might just be my eyes but looks like the pinned vine dependency is listed twice and this PR only removes one of the entries.

requirements/default.txt
```diff
@@ -1,7 +1,6 @@
 celery>=4.3.0,<5.0.0; python_version>="3.7"
 vine==1.3.0
```
bmyrtil-fsns, line 2: Shouldn't this one have been removed as well?

jrbenriquez: Yes please? Still getting `flower 0.9.7 has requirement vine==1.3.0, but you'll have vine 5.0.0 which is incompatible.`.

PR #1058, opened 2021-01-06
Broken link

areisrosa:
Add link correct for site Celery.


mher:
Thanks for the pull request! https://docs.celeryproject.org seems better

PR #1059, opened 2021-01-12
remove second instance of vine==1.3.0 (#1056)

hackslashX:
requirements.txt for some reason has duplicated vine version dependency declaration. One of them was removed in #1056, remove the second one now :(

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,5 +1,5 @@
 celery>=4.3.0,<5.0.0; python_version>="3.7"
-vine==1.3.0
+vine==5.0.0
 tornado>=5.0.0,<7.0.0; python_version>="3.5.2"
 prometheus_client==0.8.0
 humanize


hackslashX:
Yikes. Setting it to latest vine version ...

jrbenriquez:
@hackslashX Hey found this from pip

`ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

amqp 2.6.1 requires vine<5.0.0a1,>=1.1.3, but you'll have vine 5.0.0 which is incompatible.`

hackslashX:
> @hackslashX Hey found this from pip, looks like a bug in pip's resolver
> 
> ```
> 
> ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.
> 
> We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
> 
> amqp 2.6.1 requires vine<5.0.0a1,>=1.1.3, but you'll have vine 5.0.0 which is incompatible.
> ```

I just checked celery github repo. vine.five dependency was removed with celery 5.0.5. Flower currently supports up to 5.0.0 as from dependency. Updating celery dependency will probably break builds for older celery versions due to vine requirement.

hackslashX:
Yep. Installation and build works fine with celery 5.0.5. However, we need to check in with @mher to see if celery update is required at this stage. Since for now, people who really want to use flower can simply use celery 5.0.0 as per the requirements.

pabs3:
@hackslashX since vine isn't used in the flower codebase at all, shouldn't it just get removed from the `requirements/default` file instead of bumping the pinning to `5.0.0`?

mher:
Closing this PR. Vine was already removed in 7ee34cb5fc039bbeb208a54f113713edbb4809ce

PR #1073, opened 2021-02-21
Check if broker and nodes are ready before starting Flower in Docker

c-v-ya:
I've noticed that if Flower starts before the broker inside Docker it then can't connect and therefore can't provide statistics. I'm not talking here about Docker Compose way.

So here is a somewhat simple solution. If broker and nodes are ready it starts Flower. Also, checks can be disabled via environment variables. E.g. to completely skip checking broker we can add `FLOWER_SKIP_BROKER_READY=1`. And it will start Flower as it does now. Or `FLOWER_SKIP_NODE_CHECK=1` allows to just check if the broker is ready but skip checking nodes/workers.

I'm not sure how to make it better in terms of current users. In my implementation checks will run unless environment variables are provided. Or maybe it's better to enable checks by providing such variables.

I'm open for discussion and hope we'll come to a solution that suits us all.

And of course, thank you for this amazing tool!

### DIFF
--- a/wait-for-broker.sh
+++ b/wait-for-broker.sh
@@ -1,0 +1,28 @@
+cmd="$@"
+
+if [ ! -z "$FLOWER_SKIP_BROKER_READY" ]; then
+	>&2 echo "Skipping broker readiness, executing command"
+	exec $cmd
+fi
+
+while true;
+do
+	timeout 10 celery inspect ping
+	STATUS=$?
+
+	if [[ "$STATUS" == "143" ]]; then
+		>&2 echo "Broker isn't responding"
+		continue
+	fi
+
+	if [ -z "$FLOWER_SKIP_NODE_CHECK" ] && [[ "$STATUS" == "69" ]]; then
+		>&2 echo "Nodes aren't responding"
+		continue
+	fi
+
+	break
+
+done
+
+>&2 echo "Broker ready, executing command"
+exec $cmd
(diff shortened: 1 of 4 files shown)


wait-for-broker.sh
```diff
@@ -0,0 +1,28 @@
+cmd="$@"
+
+if [ ! -z "$FLOWER_SKIP_BROKER_READY" ]; then
+	>&2 echo "Skipping broker readiness, executing command"
+	exec $cmd
+fi
+
+while true;
```
nvtkaszpir, line 8: infinite loops are usually not the best idea, there should be some execution limit, like 15 tries (with optional exponential backoff.

wait-for-broker.sh
```diff
@@ -0,0 +1,28 @@
...
+		>&2 echo "Nodes aren't responding"
+		continue
+	fi
+
+	break
+
+done
+
+>&2 echo "Broker ready, executing command"
+exec $cmd
```
nvtkaszpir, line 28: please add newline at the end of the file.

nvtkaszpir (commented):
I don't believe waiting for the broker is really required, the app start should not depend on some other backend service.
Thus it would be better if flower was actually reporting empty metrics or at least basic stats of what is happeing, for example flower_status{state="connecting"} or something like it.

mher:
I think the purposed functionality should be implemented outside of flower. You can easily implement it in kubernetes or other orchestration system.

PR #1089, opened 2021-04-04
Additional Prometheus metrics.

Tomasz-Kluczkowski:
Hello,

I love that we have Prometheus metrics built into Flower but I am badly missing 2 of them: `worker online` (up) gauge and `task queuing time`. 
Both of them are easily obtainable and I do remember that queuing time was actually displayed as a graph in the `monitor` tab in flower 0.9.4 :).

So I took some small effort to bring them back :).

Please review and let me know what you think about the changes so far / if there are any more metrics that we could easily add and would be useful.

**Screenshots:**

**Task queuing time test**
Screenshot from 2021-04-04 21-16-27

**Worker online test**
Screenshot from 2021-04-04 21-15-38

**Metrics endpoint test**
Screenshot from 2021-04-04 21-10-26

I also would like to add more documentation for people about Prometheus metrics in Flower as I believe many users have no idea that they are there or how to use them (I had similar problem initially). Please let me know if this is a good idea and **what files** in the docs are best to target for that.


flower/events.py
```diff
@@ -6,28 +6,26 @@
...
-from prometheus_client import Counter as PrometheusCounter, Histogram
+from prometheus_client import Counter as PrometheusCounter, Histogram, Gauge
 
 logger = logging.getLogger(__name__)
 
 
 class PrometheusMetrics(object):
     events = PrometheusCounter('flower_events_total', "Number of events", ['worker', 'type', 'task'])
     runtime = Histogram('flower_task_runtime_seconds', "Task runtime", ['worker', 'task'])
+    queuing_time = Gauge('flower_task_queuing_time_seconds', "Task queueing time", ['worker', 'task'])
```
Tomasz-Kluczkowski, line 27: I felt that for those new metrics I prefer current, most accurate measurement rather than a histogram.

flower/events.py
```diff
@@ -39,31 +37,46 @@ def __init__(self, *args, **kwargs):
         self.metrics = PrometheusMetrics()
 
     def event(self, event):
+        # Save the event
+        super(EventsState, self).event(event)
```
Tomasz-Kluczkowski, line 41: I moved this up to have current state when doing further checks in this method.

flower/events.py
```diff
@@ -39,31 +37,46 @@ def __init__(self, *args, **kwargs):
...
             if not task_name and task_id in self.tasks:
-                task_name = self.tasks[task_id].name or ''
+                task_name = task.name or ''
             self.metrics.events.labels(worker_name, event_type, task_name).inc()
 
             runtime = event.get('runtime', 0)
             if runtime:
                 self.metrics.runtime.labels(worker_name, task_name).observe(runtime)
 
+            task_started = task.started
```
Tomasz-Kluczkowski, line 60: When playing with starting/stopping celery worker and starting/stopping Flower there could be a situation that we have a task-started registered without task-received so we need to confirm both task received and started are present on the task.

flower/events.py
```diff
@@ -39,31 +37,46 @@ def __init__(self, *args, **kwargs):
...
+            task_received = task.received
```
Tomasz-Kluczkowski, line 61: I have followed what is done so far in this method and did not split it into sub-methods/sub-handlers as I can see the `Receiver` object takes only 1 handler and I guess for a good reason.

flower/events.py
mher: `task-received - task-started` is the time task spends in a worker waiting to run. The time task spends in broker will not be included, so `queuing_time` name is a bit misleading.

mher:
Thanks for the pull request @Tomasz-Kluczkowski! You can add docs to https://github.com/mher/flower/tree/master/docs. I'd also suggest to add sample Prometheus/Grafana configs to https://github.com/mher/flower/tree/master/examples

flower/events.py
Tomasz-Kluczkowski: Yeah indeed. The name is not ideal but I guess it is a good approximation. In general I expect my broker to act nearly instantly and it has its own metrics so I feel covered there. My inspiration was code in flower 0.9.4:

`queue_time += task.started - task.received`, but I am all for 'explicit is better than implicit', so what about: `flower_task_queuing_time_at_worker_seconds` or `flower_task_waiting_time_at_worker` ? 

Personally I like the word `queuing` in there.

I also noticed that we were avoiding calculating this `queue time` if task had an `eta` set on it as it makes no sense (it will be waiting at the worker then). I could for those cases calculate queuing time as: `task.started - task.eta` showing what was the delay in actually starting the task past the initial eta value or skip completely as it was done in Flower 0.9.4 where we have:

```
    if task.eta is not None:
        continue
```

What do you think is better?

Tomasz-Kluczkowski: I have changed to: `flower_task_queuing_time_at_worker_seconds`.

Tomasz-Kluczkowski:
Add alerts examples, grafana dashboard, documentation.

Screenshot from 2021-04-05 21-40-45

flower/events.py
Tomasz-Kluczkowski: The question still stands about the `eta`. I never used it (prefer cron to schedule stuff) but if you think it's best to skip computing queuing time for eta tasks please let me know.

mher: I think it makes sense to filter out the tasks with eta

Tomasz-Kluczkowski: I will follow up with some handling of the ETA then similar to what Flower did in 0.9.4.

PR #1093, opened 2021-04-10
Prometheus metric improvements

Tomasz-Kluczkowski:
Hi,

I add further, small improvements to the Prometheus metrics.

- Make sure queuing time is not calculated if task has an ETA set.
- Add new metric: number of currently executing tasks at worker. (this one turns out to be super missing at my workplace again :P )

Replace the grafana dashboard example to contain the new metric.
Update screenshot.


tests/unit/views/test_monitor.py
```diff
@@ -20,11 +21,18 @@ def test_metrics(self):
...
-        events = [Event('worker-online', hostname=worker_name)]
+        events = [Event('worker-online', hostname=worker_name), Event('worker-heartbeat', hostname=worker_name, active=1)]
         events += task_succeeded_events(worker=worker_name, name=task_name, id='123')
+
+        task_received = time.time()
+        task_started = task_received + 3
         for i, e in enumerate(events):
             e['clock'] = i
             e['local_received'] = time.time()
+            if e['type'] == 'task-received':
```
Tomasz-Kluczkowski, line 32: Added this to the test to be able to assert on the actual queuing time.

Tomasz-Kluczkowski:
Looks like there is an issue with the test for ETA. It works fine when run individually so I suspect some problem with test isolation.

mher:
I think the problem is in prometheus metrics registry. Prometheus python client uses global registry.

mher:
https://stackoverflow.com/questions/62488819/prometheus-client-python-how-to-assign-new-registry-without-restart

Tomasz-Kluczkowski:
Thx, hmm yes it must be it as all the other objects are fresh instances.

My solution if I cannot isolate in any other way will be to simply test for different set of labels (task name / worker) and assert on them so that even if we have the old data in registry the new should be what I test for (bit of a copout but if it is awkward to isolate in another way then it has to do).

I will come back to it later in the week.

Sorry but Diablo 2 Ressurected technical alpha won over my coding passion this weekend ;).



On 12 Apr 2021, at 00:17, mher ***@***.***> wrote:

﻿

I think the problem is in prometheus metrics registry. Prometheus python client uses global registry.

Tomasz-Kluczkowski:
> https://stackoverflow.com/questions/62488819/prometheus-client-python-how-to-assign-new-registry-without-restart

I did indeed try this and it creates even worse state where after the registry is reset the metrics come as an empty string. I tried placing the collector reset code in the get_app method as well as set_up and teardown - same effect. Therefore I chose to use different labels.

If you think it is a big issue (well it was like that all the time ) then maybe we should instead of using the implicit REGISTRY, inject it into the PrometheusMetrics class. This way we can use any registry we want, and in tests inject a fresh one. I would then maybe also inject PrometheusMetrics into EventState. But for simplicity it is not needed so not sure. Let me know your opinion please.

Tomasz-Kluczkowski:
This now works. Awaiting your opinion Mher about the tests.

mher:
Looks good, thank you!

PR #1094, opened 2021-04-19
Remove the favicon and Apple touch icons

pabs3:
The touch icons are from Bootstrap and the favicon is from Celery but
Flower should use its own branding instead of Bootstrap/Celery branding.

New branding for Flower can be added later.


pabs3:
Um, this doesn't seem to have been merged, why was it closed?

pabs3:
I have updated this patch and resubmitted it as #1282 and I hope it can be included this time.

Issue #1029, opened 2020-09-25
Upgrading to Celery 5.0.0 breaks flower due to change in celery.bin.base module.


mbayabo:
If you run flower with Celery 5.0.0 or if you use the docker image, it will say it cannot import "Command".

1. ```pip install celery==5.0.0```
THEN
2. ```flower --version```
OR
3. ```flower -A proj --broker=redis://redis```
OR
4. ```flower```

![image](https://user-images.githubusercontent.com/18492353/94216329-f3abed80-fe93-11ea-887d-e411eb48694e.png)

The Command class is no longer part of the base module:
[https://docs.celeryproject.org/en/v5.0.0/reference/celery.bin.base.html](https://docs.celeryproject.org/en/v5.0.0/reference/celery.bin.base.html)

Which means it can no longer import and inherit **Command**:

```command.py
from __future__ import absolute_import
from __future__ import print_function

import os
import sys
import atexit
import signal
import logging

from pprint import pformat

from logging import NullHandler

from tornado.options import options
from tornado.options import parse_command_line, parse_config_file
from tornado.log import enable_pretty_logging
from celery.bin.base import Command

from . import __version__
from .app import Flower
from .urls import settings
from .utils import abs_path, prepend_url
from .options import DEFAULT_CONFIG_FILE, default_options


logger = logging.getLogger(__name__)


class FlowerCommand(Command):
    # ...
```

ifokeev:
Same problem here

eliorc:
Same problem here as well

kdiri:
Same problem here also.

packetflow:
Same for me, needed downgrade to 4.4.7

zufj:
Same here, also had to downgrade sadly

cesarFrias:
Same for me, dowgraded celery to 4.4.7 as @packetflow

PedroCaicedoDev:
Same problem for me

mbayabo:
Hello All,

A work around if you don't want to downgrade all of celery is to use a docker container for flower:

```
  flower:
    image: mher/flower
    command: ["--app your_app", "--url_prefix=flower", "--broker=redis://redis"]
    ports:
      - 5555:5555
    restart: unless-stopped
```

waldner:
Why was the issue closed? Is this resolved?

sogloarcadius:
This issue should not be closed because not solved

Regards

mandarvaze:
[...]

PR #1101, opened 2021-05-08
Improvements to the celery-flower-prometheus-grafana integration documentation

Tomasz-Kluczkowski:
I have been asked for a video version of full integration from celery to flower to prometheus to grafana by members of the community.
This made me think that to get this all set up is not so trivial and better manual could help :).

I don't do video but I like good manuals...

So here we go - step by step docs with lots of pics to make it easy.

I deleted all my set up and re-created it using them and all worked so hopefully others can do same and make Flower even more useful.

BTW @mher is there anything I can help with to get the newest master released?


mher:
Looks good @Tomasz-Kluczkowski . Thanks for the pull request!

> 
> BTW @mher is there anything I can help with to get the newest master released?

There are two issues with the master version:

1. `flower` command requires subcommand. Running flower with `flower` is not working anymore. `flower flower` should be used as a workaround. I'd like to add a default subcommand for not breaking backward compatibility. Scripts running flower will break with the new release.
2. There are bug reports in #1029 issue. They need to be verified and fixed if they are legit reports.

Issue #1097, opened 2021-04-23
Flower does not collect shared tasks within Django project


Real-Gecko:
**Describe the bug**
I have a Django project with lot of shared tasks scattered across `tasks.py` modules. With celery 4 Flower was able to collect them all and show 'em in tasks tab, but after upgrading to Celery 5.0.5 I see nothing there:
![image](https://user-images.githubusercontent.com/2231969/115832564-da245180-a434-11eb-9ff1-c8bfb4ffa6cd.png)

I also noticed that when I run `celery -A myproj flower` in console list of registered tasks
```shell
[I 210423 10:06:31 command:142] Registered tasks:
```

does not shall all those `@shared_tasks`s

**To Reproduce**
1. Create a Django project with an app and add usual `shared_task` to the app.
2. Start celery as usual `celery -A proj worker -E -B -l DEBUG`
3. Notice the list of tasks registered by Celery, it contains `shared_task`s
4. Start Flower `celery -A proj flower`
5. Notice that no `shared_task`s is registered.

**Expected behavior**
See the list of all tasks with statuses.

**System information**
Python 3.8.9
Celery 5.0.5
Flower https://github.com/mher/flower/commit/780bc43ce0c9545b25b3cee987fe31a5a465e0fd

mher:
Fixed by #1108

PR #1108, opened 2021-05-17
Fix autodiscovery

wblxyxolbkhv:
Fixes this [issue](https://github.com/mher/flower/issues/1097)

Root cause of this issue is Celery autodiscovery feature is not working:
- celery_app.autodiscover_tasks method (without force flag) only connects to celery import_modules signal;
- import_modules signal sent only after calling import_default_modules() method. Other common subcommands for celery (like [worker, beat](https://github.com/celery/celery/blob/master/celery/loaders/base.py#L111), [report](https://github.com/celery/celery/blob/master/celery/bin/celery.py#L157), [shell](https://github.com/celery/celery/blob/master/celery/bin/shell.py#L137)) call this method manually.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -95,3 +95,4 @@ Simon Gurcke
 Jason Held
 Lukas Matta
 Tomasz Kluczkowski
+Alexey Nikitenko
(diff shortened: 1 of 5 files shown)


wblxyxolbkhv:
@mher could you please take a look?

mher:
@wblxyxolbkhv thanks for the pull request. Can you resolve the conflict in tests?

wblxyxolbkhv:
@mher Done!

PR #1110, opened 2021-05-23
Update command line instructions and invocation

Tomasz-Kluczkowski:
A lot of doc updates removing any instructions showing usage of Flower as `flower [args]`.
Users should now always use `celery [celery args] flower [flower args] as per Celery 5 instructions.

Removed flower command line alias from `setup.py`.
Add a warning when instantiating Flower if user specified celery args in the flower part.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -57,13 +57,33 @@ Development version: ::
 Usage
 -----
 
-Launch the server and open http://localhost:5555: ::
+**Important** Please note that from version 1.0.1 Flower supports Celery 5.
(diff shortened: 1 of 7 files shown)


Tomasz-Kluczkowski:
@mher This is an initial version of improvements that hopefully can get newest Flower finally released :)
I tried to improve docs as much as possible.
Please check carefully if I made any mistakes, I hope this will get people to use Flower correctly.

I need to reinstall this locally with the new `setup.py` without `flower` in `entry_points` to make sure one can no longer use `flower` straight from command line. I have not played a lot with `setup.py` in the past. Do I just run `pip install flower develop` or something like that?

PR #1116, opened 2021-06-18
Add purge_offline_workers option to configuration docs

fjcapdevila:
I've found `purge_offline_workers` in the code but it was not in the documentation.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -222,6 +222,13 @@ port
 
 Run the http server on a given port (by default, `port=5555`)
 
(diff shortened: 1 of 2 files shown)


fjcapdevila:
Should I update this line too? Looks like the test is exclude 4 existing options on purpose.
https://github.com/mher/flower/blob/master/tests/unit/test_command.py#L120

mher:
`purge_offline_workers` is in the docs https://github.com/mher/flower/blob/master/docs/config.rst#id35

PR #1117, opened 2021-06-20
Add number of tasks queuing at worker Prometheus metric

Tomasz-Kluczkowski:
@mher Sorry to add to your workload.

In this PR I add a new metric that was requested in the community and also needed at my workplace - number of tasks queuing up at a worker.

With this metric and the queuing time one, the devops engineer should be able to make a decision on hardware/VM requirements easier and also diagnosing any Celery task soft/hard timeout issues for tasks should be easier.

I also fix the issue with task queuing time getting stuck at last recorder value if no other tasks of that type ever run on a given worker - now we just reset the queuing time to zero after the task succeeded/failed.

Docs updated accordingly (just a row in a table)

Please let me know if you have any comments.


mher:
Thanks for the pull request @Tomasz-Kluczkowski 

I think `flower_number_of_tasks_queuing_at_worker` is an ambiguous name. I'd suggest to follow celery's terminology (see https://docs.celeryproject.org/en/stable/userguide/optimizing.html#prefetch-limits). Something like `flower_worker_prefetched_tasks`.

mher:
`flower_task_queuing_time_at_worker_seconds` also can become `flower_task_prefetch_time_seconds`.

Tomasz-Kluczkowski:
> Thanks for the pull request @Tomasz-Kluczkowski
> 
> I think `flower_number_of_tasks_queuing_at_worker` is an ambiguous name. I'd suggest to follow celery's terminology (see https://docs.celeryproject.org/en/stable/userguide/optimizing.html#prefetch-limits). Something like `flower_worker_prefetched_tasks`.

both your comments make total sense and that actually is what is measured. I will adjust naming all over the place accordingly.

Tomasz-Kluczkowski:
@mher I have updated the naming of the metrics to use `prefetch` as Celery guys do - this meant docs, grafana stuff and prometheus also had to be amended.
Updated grafana dashboard to contain new metric and exported to use Prometheus data source by default and showed that in the manual.

Please let me know if there are any outstanding issues.

PR #1123, opened 2021-07-07
Fix Dockerfile CMD

GabrielBG99:
The dockerfile CMD was not working because it was treating the "celery flower" as a single command


mher:
cc @Tomasz-Kluczkowski

PR #1133, opened 2021-08-03
Add Auth0 login handler

itssimon:
Allows using Auth0 as an authentication provider. The mechanism is exactly like for Okta, so all it took was to update the authorize, access token and userinfo URLs.

The base URL is configurated via the environment variable `FLOWER_OAUTH2_AUTH0_BASE_URL`.


itssimon:
@mher, any chance you could have a look at this and merge?

mher:
Flower supports the use of custom authentication providers, and there is no need to merge them into the project. Simply create your own authentication class and use it with the --auth-provider option.

PR #1149, opened 2021-09-26
making PrometheusMetrics class singleton

alienr:
@Tomasz-Kluczkowski I tried adding only __init__ method but there were multiple initialisation of this class and because of that it was giving metric with name already exist. so I made it singleton. basically it is extension of your idea. 

if you have any other idea, pls let me know

### DIFF
--- a/flower/events.py
+++ b/flower/events.py
@@ -24,30 +24,39 @@ logger = logging.getLogger(__name__)
 
 
 class PrometheusMetrics(object):
-    events = PrometheusCounter('flower_events_total', "Number of events", ['worker', 'type', 'task'])
+    __instance = None
 
-    runtime = Histogram(
-        'flower_task_runtime_seconds',
-        "Task runtime",
-        ['worker', 'task'],
-        buckets=options.task_runtime_metric_buckets
-    )
-    prefetch_time = Gauge(
-        'flower_task_prefetch_time_seconds',
-        "The time the task spent waiting at the celery worker to be executed.",
-        ['worker', 'task']
-    )
-    number_of_prefetched_tasks = Gauge(
-        'flower_worker_prefetched_tasks',
-        'Number of tasks of given type prefetched at a worker',
-        ['worker', 'task']
-    )
-    worker_online = Gauge('flower_worker_online', "Worker online status", ['worker'])
-    worker_number_of_currently_executing_tasks = Gauge(
-        'flower_worker_number_of_currently_executing_tasks',
-        "Number of tasks currently executing at a worker",
-        ['worker']
-    )
+    @staticmethod
+    def get_instance():
+        if PrometheusMetrics.__instance is None:
+            PrometheusMetrics.__instance = PrometheusMetrics()
+        return PrometheusMetrics.__instance
+
+    def __init__(self):
+        self.events = PrometheusCounter('flower_events_total', "Number of events", ['worker', 'type', 'task'])
+
+        self.runtime = Histogram(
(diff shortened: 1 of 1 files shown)


Tomasz-Kluczkowski:
Ok, I have found more time to look at  this problem and tried the initial idea of just simple init method in the `PrometheusMetrics` class to observe what is actually reporting the duplicate time series problem.

What actually happens is the **tests** are reporting that error, not actual running flower application - since there is no proper isolation (the prometheus registry is global). I have been hit by this before and it was annoying but I have not fixed it.

The easiest way to actually isolate tests, and also the proper way to create metrics as per official docs is to have it as module level variables. I tried that, and this fixes the duplicated time series issue, **but** since we use then a module level variable, it gets its buckets set in stone too early, before options are modified , same as when we used class variables. So **not good** for us.

So for our use case, **we have to delay instantiation of the metrics** and make sure they work ok in tests.
Therefore I would say your approach is ok.

flower/events.py
```diff
@@ -24,30 +24,39 @@
...
-    )
-    worker_online = Gauge('flower_worker_online', "Worker online status", ['worker'])
-    worker_number_of_currently_executing_tasks = Gauge(
-        'flower_worker_number_of_currently_executing_tasks',
-        "Number of tasks currently executing at a worker",
-        ['worker']
-    )
+    __instance = None
+
+    @staticmethod
```
Tomasz-Kluczkowski, line 29: I think this should be a class method. I don't expect us to inherit from this class, but then it in principle would work for any class based on this one.

```python
    @classmethod
    def get_instance(cls):
        if cls.__instance is None:
            cls.__instance = cls()
        return cls.__instance
```

an alternative I guess is to have a global variable which gets only set once when we need it in EventsState and then use a factory method. @mher any preferrences?

alienr: @Tomasz-Kluczkowski I think second idea is enough. I have implemented the same. please have a look

Tomasz-Kluczkowski:
Ok, just had another quick look - the error comes from `humanize` package for which we don't specify version in `default.txt` and I assume newest one gets installed and it no longer has `humanize.VERSION` attribute.

```python
module 'humanize' has no attribute 'VERSION' <- I just printed the exception in `bug_report`
______________________________________________________________________________________________________________________________________________________________________________________________ BugreportTests.test_with_app ______________________________________________________________________________________________________________________________________________________________________________________________

self = <tests.unit.utils.test_utils.BugreportTests testMethod=test_with_app>

    def test_with_app(self):
        app = Celery()
        report = bugreport(app)
>       self.assertFalse('Unknown Celery version' in report)
E       AssertionError: True is not false
```

@mher this function: `bugreport` reports unknown celery version when a number of exceptions is raised and that makes no sense. I understand why we have it, but I think we are better off fixing the package versions that Flower uses in all requirement files rather than not specifying them and when they get a new release stuff starts breaking on any merge to master and we rely on this bug_report to let us know that some dependency no longer will work the way we need it to.

Would you agree on fixing package versions in all requirements (try to fix them at the newest current versions)?
We can then use dependabot (or whatever this feature is called in github) to notify us/create an auto PR with new dependency version any time they get a new release...

As a fix to this build failing I will see if `humanize` provides some other way of getting a version and use it inside `bug_report`.

I also thing we should change the message in `bug_report` to show actual exception message if ImportError/AttributeError happens plus add a message: 'Have you installed correct versions of Flower's dependencies?'

thoughts?

Tomasz-Kluczkowski:
Fix here: #1156

PR #1155, opened 2021-10-16
Use assertEqual instead of assertEquals for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268


mher:
Already fixed in the master

PR #1176, opened 2021-12-11
Remove deprecated alias.

ldav1s:
I noticed this when running `tox`.  Should be OK to remove since you specify your minimum target Python version as 3.6.


ldav1s:
I guess I should have looked at #1155 first.

Issue #1140, opened 2021-08-27
Docker tag for 1.0.0 release


daveisfera:
**Is your feature request related to a problem? Please describe.**
1.0.0 was [released](https://github.com/mher/flower/releases/tag/v1.0.0) but there's not a docker tag for it

**Describe the solution you'd like**
Add tag so 1.0.0 can be used in a docker image

cfarrend:
This would be great for us! We used to use the `latest` tag but unfortunately it broke a while back and we needed to version lock.

cfarrend:
For anyone who is watching this, I have built the `1.0.0` release as `cfarrend/flower:v1.0.0`. I'm basing it off the same code except with one slight Dockerfile modification. If it works out then we'll be using it anyway so it will be maintained to the point of supporting new versions as they come, of course I'd prefer not to. I've also built the latest push to `master` as `cfarrend/flower:master`. This should update automatically as pushes are added hopefully.

See here:
https://github.com/mher/flower/compare/v1.0.0...cfarrend:v1.0.0

I'd make a PR here but I'm not sure of the build process to be honest. Possibly done inside DockerHub itself?

cardoe:
I've created https://github.com/mher/flower/pull/1181 which will switch this project to GitHub Actions for container building. This will bring back master building to latest on merges and also start building release tags again. Docker Hub Automated Builds are disabled by default now which is why this has stopped working.

cardoe:
I've followed up on PR #1181 but unfortunately something isn't working yet so we can't close this one out.

mher:
Fixed in bfca6860ef7fc88a403ae8550f7ae669a7709a63

Issue #1126, opened 2021-07-17
Docker tag "latest" isn't latest


Hultner:
## Background
I started getting some problems with flower when I upgraded to Celery 5, had to fiddle around quite a bit until I found this configuration in my docker-compose to work
```yaml
  flower:
    image: mher/flower:latest
    networks:
      - ${PUBLIC_NETWORK?Variable not set}
      - default
    depends_on: 
      - queue
      - pdf-celeryworker
      - celeryworker
    env_file:
      - .env
    command: celery flower
      # The ones below did not work anymore
      # - "--broker=amqp://[REDACTED]:5672//"
      # - "celery --broker=amqp://[REDACTED]:5672// flower"
    environment: 
      CELERY_BROKER_URL: "amqp://[REDACTED]:5672//"
      CELERY_RESULT_BACKEND: "rpc://[REDACTED]//"
    deploy:
      labels:
        [REDACTED]
```

After some more investigation that the `latest`-tag pushed at `"2021-07-07T01:41:23.237381Z"` is actually a previous version, 0.9.5. 

This can be verified by running `flower --version` inside of the container.
```
docker-compose exec flower flower --version
0.9.5
4.4.7 (cliffs)
```

From the releases on GitHub I expected latest to be version 1.0, but no tag at version 1 seem to exists.

rightisleft:
I saw that as well

h55nick:
Just ran into this as well.

rightisleft:
I know docker hub is charging for auto builds soon - maybe related? 

It's pretty straightforward to use flower as a pip library, but debugging this did lose me a few hours of my life.

abdrat:
I had the same issue. For the benefit of google searchers flower in docker was throwing "exited with code 127" error on the :latest tag.

cardoe:
I've created https://github.com/mher/flower/pull/1181 which should fix this.

PR #1181, opened 2021-12-31
use GitHub Actions to build Docker container

cardoe:
Start using GitHub Actions to build the Docker container for this
project. Builds all tags and master as the latest container tag. Builds
for amd64 and arm64.


cardoe:
This will require you to to add two secrets `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` under Settings -> Secrets -> Environment Secrets. The later is expected to be a Docker Hub Personal Access Token. See https://github.com/docker/login-action#docker-hub for more information as this is the action that is used.

The reason for this PR is that the project current uses Docker Hub's Automated Build to generate the containers however per https://www.docker.com/blog/changes-to-docker-hub-autobuilds/ they have disabled auto builds by default unless the project signs up as an official Open Source project, which this project hasn't done. Additionally the Automated Build environment is a bit limited for things like multi-platform builds (which this PR uses) and lots of folks have started to switch to GitHub Actions for container building.

cardoe:
Note: This PR will *fail* until the author sets up the Secrets appropriately.

cardoe:
fixes #1126 fixes #1140

mher:
@cardoe login action is failing despite adding `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`

cardoe:
https://github.com/docker/login-action#docker-hub are the docs for the action. It appears they've moved to v2 so we can try going to v2, otherwise I can only guess its a creds issue.

scottd018:
Also, @mher is there a chance you can push a new image for version `v1.1.0` to the `mher` Docker Hub repo when you get a chance?  This would unblock me from using flower for my application.  Right now, the latest one in Docker Hub has several vulnerabilities which won't allow me to use it.  I have built and scanned off of master and it looks good.  The challenge being that I don't want to push and maintain the forked copy (because we all know how that goes...ha).

Apologies for this one-off request, as we have some tight security controls that I'm working around.  Much appreciated for being responsive on this one!

mher:
> Functionally there's no difference between v1 and v2. v2 just requires a newer GitHub Actions Runner version so they treated that a semver change. Only guess is its a cred issue. You made a Personal Access Token https://docs.docker.com/docker-hub/access-tokens/ that had permissions to push to `mher/flower`?

@cardoe yes, the token has `Read, Write, Delete` scope. I'm able to `docker login` from the command line using the same credentials.

mher:
Moving secrets from `Environment secrets`to `Repository secrets` solved the issue

scottd018:
> I don't believe you are correct. Those triggers are already done. We additionally include PRs so that we attempt to build the Docker container on PRs. If you look at the login action
> 
> https://github.com/mher/flower/blob/d91fbbf338044c0668cf053ccf1c5a33c6454f0e/.github/workflows/docker.yml#L34
> 
> it explicitly skips for PRs. The failed run didn't happen on the PR, the failed run happened in master. You can even see the follow on PR built here https://github.com/mher/flower/actions/runs/2634395754 successfully.

Ah, yeah, my bad.  Looked at that wrong.

Regardless, looks like you have it solved and working.  I see the `master` tag in DockerHub now, so I can use that one until a tagged release is present.  Thanks again @mher and @cardoe !

cardoe:
Awesome. Thanks for the follow up @mher didn’t even think about the difference between the two. I’ll make a follow up PR to have master go to latest.

PR #1200, opened 2022-03-21
Fixing `CORS` support 

### LINKED ISSUE #1199: TypeError: options() takes 1 positional argument but 2 were given
**Describe the bug**
When making a cross-site request to the flower API with `basic_auth` on `/api/task/result/<id>`, tornado is raising a TypeError because:

- No `POST` methos is allowed to make cross-site request to this endpoint
- `def options(self):` _takes 1 positional argument but 2 were given_
- Since we are sending an `authorization` header with the current credentials the browser is generating a preflight request before the actual request to know what headers and method we can actually use

**To Reproduce**
1. Launch flower with basic authentication using `celery -A celery_worker.celery flower --basic_auth=user1:pass1` eg.
2. Make a request to `/api/task/result/<id>` including the Authorization header, for example:

```javascript
fetch("http://localhost:5555/api/task/result/12f4703c-0234-4da6-9170-5cee63d36455", {
  "headers": {
    "authorization": "Basic S1kZ6OFpjNGoWEQ3Yjk5g="
  },
  "referrer": "http://localhost:3000/",
  "method": "GET"
});
```

**Expected behavior**
Successfully send and process the authenticated request to Flower.

**Screenshots**
image

image

**System information**

It returns:
 `Unknown Celery version`

however
```
flower   -> flower:1.0.1 tornado:6.1 humanize:3.12.0
software -> celery:5.1.2 (dawn-chorus) kombu:5.2.4 py:3.9.9
            billiard:3.6.4.0 redis:4.1.4
...
```

dave-flr:
### What?

- Allow `POST`, `GET` and `PATH` Methods 
- Allow `access-control-allow-origin,authorization,content-type` Headers
- Added `*args` and `*kwargs` to `def options(self)` signature

Why?
- being able to send those methods to a basic authenticated flower instance 
- allow sending the Authorization headers with the base-64 encoded user and password as well as other frequently used headers such as Content-Type
- prevent and fix #1199 to happen according to tornado [documentation](https://www.tornadoweb.org/en/stable/web.html?highlight=RequestHandler#tornado.web.RequestHandler.options)

### DIFF
--- a/flower/views/__init__.py
+++ b/flower/views/__init__.py
@@ -17,7 +17,8 @@ logger = logging.getLogger(__name__)
 class BaseHandler(tornado.web.RequestHandler):
     def set_default_headers(self):
         self.set_header("Access-Control-Allow-Origin", "*")
-        self.set_header("Access-Control-Allow-Headers", "x-requested-with")
+        self.set_header("Access-Control-Allow-Headers",
+                        "x-requested-with,access-control-allow-origin,authorization,content-type")
         self.set_header('Access-Control-Allow-Methods',
                         ' PUT, DELETE, OPTIONS, POST, GET, PATCH')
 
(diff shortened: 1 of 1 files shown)


dave-flr:
I'm already using this fix in [this](https://github.com/dave-flr/flower) fork  to move around from this bug since https://github.com/mher/flower/pull/994 is not working properly for me

PR #1205, opened 2022-04-08
The old templates are not consistent with newer versions of Django wh…

Jokezor:
# What
The error and 404 templates are not consistent with newer versions of Django which makes 404 pages crash when running flower through Django.

When an 404 is raised in Django it finds the templates in flower and fails on the templates.
This is the case I'm trying to solve.

# Changes
- Changed from {% end %} to {% endif %} if there was a conditional statement.
- Changed from {% end %} to {% endblock %} if there was a block.
- Changed from {% end %} to {% endfor %} if there was a for loop.

### DIFF
--- a/flower/templates/404.html
+++ b/flower/templates/404.html
@@ -7,7 +7,7 @@
                 {{ message }}
             {% else %}
                 Error, page not found
-            {% end %}
+            {% endif %}
         </p>
     </div>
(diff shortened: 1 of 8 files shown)


Jokezor:
Closing since issue was with the setup and not flower itself.

Issue #1209, opened 2022-04-26
Security Contact for Flower


tprynn:
I've attempted to reach out to disclose security issues in Flower by emailing the owner of the repo (@gmail.com) as well as contacting other members of the Celery project. I haven't been able to get a response as of yet. Is there any maintainer of this repo who can contact me (tannerprynn@gmail.com) in order to disclose these issues? I also have a PR to fix the issues but would prefer to review the PR on a private repo before making it public.

tprynn:
In the absence of a security contact for the project, I plan to disclose the issue (with a PR fix and in a blog post) on May 17 (two weeks from now).

@mher please respond if you're still maintaining the project! Thanks

tprynn:
I've delayed the disclosure for a short time to allow a project which depends on flower to make a new release.

PR #1216, opened 2022-05-26
Security fixes

tprynn:
Ref #1209, #1217 and <https://tprynn.github.io/2022/05/26/flower-vulns.html>

* Disable API by default if authentication is not configured
* Update --auth flag handling to fix email validation bypasses
* Document and test config flag changes

### DIFF
--- a/docs/auth.rst
+++ b/docs/auth.rst
@@ -6,6 +6,10 @@ Protecting your Flower instance from unwarranted access is important
 if it runs in an untrusted environment. Below, we outline the various
 forms of authentication supported by Flower.
 
+By default, Flower's APIs are not enabled unless authentication is configured.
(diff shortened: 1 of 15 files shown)


tprynn:
It looks like a PR to add full CORS support was added in the meantime. I would recommend removing that PR as it significantly increases the exploitability of CSRF attacks against Flower.

tprynn:
I've resolved the merge conflict relating to the CORS policy by setting the dangerous and unsafe CORS policy behind the same flag (`dangerous_allow_unauth_api`). Heads up @dave-flr for reference, or you have a different solution.

caleb15:
@auvipy @thedrow this could affect a large number of celery users. @mher FYI.

auvipy:
this package is not maintained by any active core developers or we don't have commit access to this.

PR #1225, opened 2022-07-01
Diable API if auth is enabled

mher:
- Disable API if authentication is enabled
- Deprecate WebSocket API


Tudyx:
Please could you indicate how  to use the API and still have authentication?
My use case is i want to give access to the web monitoring to a limited number of persons and  also use  the API to automate some task.

WillKoehrsen:
The API disabling seems like it may be a misunderstanding of the recommendation from the [blog post](https://tprynn.github.io/2022/05/26/flower-vulns.html) that identified security issues.

That blog post states:
> Disable the API unless authentication is configured for Flower

This PR disables the API **if** authentication is configured which is the opposite of the recommendation.

PR #1227, opened 2022-07-09
Security improvements

mher:
- Limit `auth` option to simple regex
- By default set random value for `cookie_secret` option
- Warn about running without authentication
- Set CORS headers only if authentication is not configured

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -65,6 +65,8 @@ auth
 ~~~~
 
 Enables authentication. `auth` is a regexp of emails to grant access.
+For security reasons `auth` only supports a basic regex syntax: single email (`user@example.com`),
+wildcard (`.*@example.com`) or list of emails separated by pipes (`one@example.com|two@example.com`).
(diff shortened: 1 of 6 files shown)


tprynn:
Hey, checking in on this, sorry about any delay in response - been busy since I just started a new job.

`cookie_secret` isn't an effective prevention for CSRF attacks. The essential innovation / interesting piece of a CSRF attack is that it abuses how web browsers handle cross-site requests. Basically, browsers will automatically send the user's cookie with most types of cross-site requests. Thus, cookies alone can't prevent a CSRF attack. There are a number of options to prevent CSRF but unfortunately none of them can realistically be implemented without making a breaking change to at least some portion of existing clients.

At line 21 of flower/views/auth.py, I'd recommend using `re.escape` rather than the manual escaping you're doing now, and replacing the `.*` wildcard with the character class `[A-Za-z0-9!#$%&'*+/=?^_`{|}~.\-]+`.

PR #1233, opened 2022-08-13
Set proper default for RabbitMQ Management API URL

### LINKED ISSUE #1232: Improper default for RabbitMQ Management API URL
Hi,

It seems that from PR #896 (released with 0.9.4) broker_api (http_api) defaults to the same port of the broker (broker_url) inherited from Celery, in addition to the same user, address, etc., instead of continuing to hardcode the port as 15672.

https://github.com/mher/flower/blob/30b5a975b763843bc7ff27d5515f33f97f986bce/flower/utils/broker.py#L52-L56

vs.

https://github.com/mher/flower/blob/95e9f84d3aefa9de2a9a9676e16ce04c8bd3ab1b/flower/utils/broker.py#L58-L60

This seems counterproductive and appears to have broken several configurations in the wild.

RabbitMQ does not appear to support having the Management plugin listening on the same port as the AMQP listener, so there should not be a use case for this default.

On the other hand, using the same port as the default Management plugin listener (15672) would allow it to work without any additional Flower specific configuration, as it did in the past.

skodde:
Closes #1232

The code was changed in the course of the 0.9 release to use the same
port as the broker for the broker_api, which doesn't work in any
scenario.

Let's revert it back to the default port of 15672.

### DIFF
--- a/flower/utils/broker.py
+++ b/flower/utils/broker.py
@@ -50,7 +50,7 @@ class RabbitMQ(BrokerBase):
         self.password = self.password or 'guest'
 
         if not http_api:
-            http_api = "http://{username}:{password}@{host}:{port}/api/{vhost}".format(
+            http_api = "http://{username}:{password}@{host}:15672/api/{vhost}".format(
                 username=self.username, password=self.password,
                 host=self.host, port=self.port, vhost=self.vhost
             )


skodde:
Replaced by https://github.com/mher/flower/commit/9089a7bf1ede4613ee6020dfb6a47eff82b1e2dd

PR #1243, opened 2022-09-08
Fixes #1242 - Ask for safe args strings when calling inspect.active

### LINKED ISSUE #1242: Worker List broken if any non-JSON serialisable attributes are present
**Describe the bug**
When a task with non JSON serializable arguments is running in Celery and `/api/workers` is called, it returns a status code of 500.

The reason why this is happening is that when inspect request is issued to celery to get active worker, task, etc https://github.com/mher/flower/blob/master/flower/inspector.py#L23-L24, the active tasks would return tasks information with the arguments passed. If the arguments contain anything that is not JSON serializable then the request would fail. Tornado tries to serialize every input that is coming in. If any non-JSON serializable arguments are present it will fail

**To Reproduce**

1. Create a long-running task with non JSON serializable arguments.

```
# tasks.py
from celery import shared_task
from time import sleep

@shared_task
def some_task(param):
    print(f"{param} was passed")
    sleep(120)
```
2. Run the task (running it on django shell)
```
In [1]: from myapp.apps.niceapp.tasks import some_task

# calling the task with a non-JSON serializable arg
In [2]: some_task.delay(set([1, 2]))
Out[2]: <AsyncResult: ab75a9fb-061c-48a2-a4cb-e85df047e23f>
```
3. Ensure that flower is running and while the task is running, make a request to `/api/workers`

```
# I am setting refresh in order to query the latest values
$ curl --location --request GET 'http://localhost:5555/api/workers?refresh=true'
```

5. You will get 500 response code to the request.

The error log is something like this
```
  HTTPServerRequest(protocol='http', host='127.0.0.1:64559', method='GET', uri='/api/workers?refresh=1', version='HTTP/1.1', remote_ip='127.0.0.1')
    File "/Users/amitphulera/.pyenv/versions/hq_celery/lib/python3.9/site-packages/tornado/web.py", line 1713, in _execute
      result = await result
    File "/Users/amitphulera/.pyenv/versions/hq_celery/lib/python3.9/site-packages/tornado/gen.py", line 782, in run
      yielded = self.gen.send(value)
    File "/Users/amitphulera/Development/dimagi/flower/flower/api/workers.py", line 185, in get
      self.write(self.application.workers)
    File "/Users/amitphulera/.pyenv/versions/hq_celery/lib/python3.9/site-packages/tornado/web.py", line 845, in write
      chunk = escape.json_encode(chunk)
    File "/Users/amitphulera/.pyenv/versions/hq_celery/lib/python3.9/site-packages/tornado/escape.py", line 75, in json_encode
      return json.dumps(value).replace("</", "<\\/")
    File "/Users/amitphulera/.pyenv/versions/3.9.9/lib/python3.9/json/__init__.py", line 231, in dumps
      return _default_encoder.encode(obj)
    File "/Users/amitphulera/.pyenv/versions/3.9.9/lib/python3.9/json/encoder.py", line 199, in encode
      chunks = self.iterencode(o, _one_shot=True)
    File "/Users/amitphulera/.pyenv/versions/3.9.9/lib/python3.9/json/encoder.py", line 257, in iterencode
      return _iterencode(o, 0)
    File "/Users/amitphulera/.pyenv/versions/3.9.9/lib/python3.9/json/encoder.py", line 179, in default
      raise TypeError(f'Object of type {o.__class__.__name__} '
  TypeError: Object of type set is not JSON serializable
```

**Expected behavior**

It should return the information of the workers and the task running in them.

**System information**

```
flower   -> flower:1.1.0 tornado:6.2 humanize:4.3.0
software -> celery:5.2.7 (dawn-chorus) kombu:5.2.4 py:3.9.9
            billiard:3.6.4.0 py-amqp:5.1.1
platform -> system:Darwin arch:64bit
            kernel version:21.6.0 imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:disabled

deprecated_settings: None
```

AmitPhulera:
This PR fixes #1242 

Celery's `inspect.active` accepts `safe` arg which would return safe strings for the objects that are not JSON serializable. This fixes the 500 error that comes in the API calls and I guess views also use the same code, so they would be fixed too.

https://github.com/celery/celery/blob/master/celery/worker/control.py#L367

- I have added a test (32554a1518f4a9cbdb2cc5e3758cec45b67b92fd) to display the issue but later I removed it (5ca140452a76e1ec61cbfe7fff94f55c27c40804) as now celery will not return unsafe strings, so the test did not made sense.

### DIFF
--- a/flower/inspector.py
+++ b/flower/inspector.py
@@ -35,7 +35,11 @@ class Inspector(object):
 
         logger.debug('Sending %s inspect command', method)
         start = time.time()
-        result = getattr(inspect, method)()
+        result = (
+            getattr(inspect, method)()
+            if method != 'active'
+            else getattr(inspect, method)(safe=True)
+        )
         logger.debug("Inspect command %s took %.2fs to complete", method, time.time() - start)
         
         if result is None or 'error' in result:
--- a/tests/unit/api/test_workers.py
+++ b/tests/unit/api/test_workers.py
@@ -17,29 +17,11 @@ empty_inspect_response = {
     'celery@worker1': []
 }
 
-inspect_response_with_non_serializable_args = {
-    'celery@worker1': {
-        'active': {
-            'id': '1234',
-            'args': [set([1, 2])]
-        }
-    }
-}
-
 
 @mock.patch.object(Inspector, 'methods',
                    new_callable=mock.PropertyMock,
                    return_value=['inspect_method'])
 class ListWorkersTest(AsyncHTTPTestCase):
-
-    def test_non_serialisable_args_in_response(self, m_inspect):
-        celery = self._app.capp
-        celery.control.inspect = mock.Mock()
-        celery.control.inspect.return_value.inspect_method = mock.Mock(
-            return_value=inspect_response_with_non_serializable_args
-        )
-        # this will fail with a response code of 500
-        self.get('/api/workers?refresh=1')
 
     def test_refresh_cache(self, m_inspect):
         celery = self._app.capp
--- a/tests/unit/api/test_workers.py
+++ b/tests/unit/api/test_workers.py
@@ -17,11 +17,29 @@ empty_inspect_response = {
     'celery@worker1': []
 }
 
+inspect_response_with_non_serializable_args = {
+    'celery@worker1' : {
+        'active': {
+            'id': '1234',
+            'args': [set([1,2])]
+        }
+    }
+}
+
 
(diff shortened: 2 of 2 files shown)


AmitPhulera:
Hey @mher!
Is there an expected timeline for this getting released?

dannyroberts:
I'm also interested in seeing this fix make it into a release!

Issue #1231, opened 2022-08-11
API is disabled when using basic_auth


jepify:
**Describe the bug**
Is it intentional that the flower API is disabled when using basic_auth.
It is a result of the change being made to all requests [here](https://github.com/mher/flower/blob/master/flower/api/__init__.py) as a fix to the [1217](https://github.com/mher/flower/issues/1217) issue about no default auth.

Is it intentional that basic_auth dissables the API even though the disclosure mentions that basic_auth is not vulnerable?

frontierpsycho:
I believe this change was perhaps the result of a misunderstanding. @tprynn's original [blog post](https://tprynn.github.io/2022/05/26/flower-vulns.html) that disclosed the security issues that the change was supposed to fix, had this recommendation for flower:

> Disable the API unless authentication is configured for Flower

But I think this was misunderstood and in the end the API was disabled _when_ (not unless) authentication is configured. 

It seems that the creator of flower has probably been too busy to maintain it, unfortunately.

snyaggarwal:
Any update on this? When can this be merged?

jangia:
Have no idea. I haven't heard anything from the package author yet.

rajkumarvenkatasamy:
Any updates on when this issue will get fixed? Thanks!

SoundWaveX81:
Another one here trying to know if this will get fixed at some point :)

adampl:
This is stupid: "Disable the API unless authentication is configured for Flower"

If at all, the API access should be disabled by default, and enabled by a parameter. Authentication and authorization can be done on multiple levels (physical, network, HTTP, application), so functionalities should not rely on auth being enabled in the app itself.

adampl:
@alexanderaltanis  It's not rude to point out that something is stupid, when it actually is stupid. Advice from a security audit shouldn't be always taken literally, because security auditors don't care about functionality and usability and separation of responsibility. If some company exposes unprotected Flower to the Internet, then it's entirely their fault and their problem.

tprynn:
@adampl I'm the person who gave that advice initially and I'll point out that I actually [submitted a PR](https://github.com/mher/flower/pull/1216) along with my original post which does exactly what you asked for (gives you a config option to enable the API without authentication).

[...]

PR #1248, opened 2022-10-12
Require auth to use api

jangia:
As per https://github.com/mher/flower/issues/1231:

API should be disabled when no authentication is enabled. At the moment, it's vice versa. The original blog post says: "Disable the API **unless** authentication is configured for Flower".

This PR updates the code and all the API tests to pass (basic_authentication is mocked).

### DIFF
--- a/flower/api/__init__.py
+++ b/flower/api/__init__.py
@@ -3,5 +3,5 @@ from ..views import BaseHandler
 
 class BaseApiHandler(BaseHandler):
     def prepare(self):
-        if self.application.options.basic_auth or self.application.options.auth:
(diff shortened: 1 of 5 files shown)


adampl:
when is it going to be merged?

adampl:
Functionalities shouldn't depend on auth being active in the app, because access can be restricted on different levels (physical, network, HTTP, application), and the app cannot (and shouldn't) know if such auth is active or not.

mher:
In the latest master version, authentication is now required for API access. However, there is an option to use the API without authentication by setting the FLOWER_UNAUTHENTICATED_API flag.

PR #1251, opened 2022-10-28
Use SPDX license expression in project metadata

RazerM:
As a downstream user it is desirable to be able to programmatically determine the precise licenses used by our dependencies. The emerging convention for this is the [SPDX License List](https://spdx.org/licenses/). SPDX License Expressions are already used in the JavaScript (npm) and Rust (crates.io) ecosystems, for example.

In Python, there is [PEP 639](https://peps.python.org/pep-0639/) which would add a standard 'License-Expression' metadata field. The discussion around this PEP seems to have gone stale in 2021.

This merge request is in lieu of that PEP coming soon. Unless you as project maintainers think the proposed change in this PR has downsides (which you are entitled to!), I would ask that you accept this change so that it's possible for users to attempt to parse your license metadata as an SPDX License Expression. In particular, BSD is ambiguous as there are multiple BSD licenses.

If PEP 639 is accepted, this package would be ready for the change just by changing the `license` key to `license-expression` or whatever key the PEP lands on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setup(
     author='Mher Movsisyan',
     author_email='mher.movsisyan@gmail.com',
     url='https://github.com/mher/flower',
-    license='BSD',
+    license='BSD-3-Clause',
     classifiers=classifiers,
     packages=find_packages(exclude=['tests', 'tests.*']),
     install_requires=get_requirements('default.txt'),


pravitva (approved):

setup.py
```diff
@@ -50,7 +50,7 @@ def get_requirements(filename):
     author='Mher Movsisyan',
     author_email='mher.movsisyan@gmail.com',
     url='https://github.com/mher/flower',
-    license='BSD',
+    license='BSD-3-Clause',
     classifiers=classifiers,
     packages=find_packages(exclude=['tests', 'tests.*']),
```
pravitva, line 55: - [ ] - @pravitva

setup.py
```diff
@@ -50,7 +50,7 @@ def get_requirements(filename):
     author='Mher Movsisyan',
     author_email='mher.movsisyan@gmail.com',
     url='https://github.com/mher/flower',
-    license='BSD',
+    license='BSD-3-Clause',
```
pravitva, line 53:
```suggestion
    license='BSD-3-Clause',
```

setup.py
```diff
@@ -50,7 +50,7 @@ def get_requirements(filename):
...
     classifiers=classifiers,
     packages=find_packages(exclude=['tests', 'tests.*']),
     install_requires=get_requirements('default.txt'),
```
pravitva, line 56: _~~```

```~~_

pravitva (approved):

pravitva (approved):

mher:
At present, [PEP 639](https://peps.python.org/pep-0639/) has not been accepted. Therefore, it would be premature to implement any changes based on this proposal.

RazerM:
changing BSD (ambiguous) to BSD-3-Clause (unambiguous) in the field flower currently uses is unrelated to PEP 639

PR #1258, opened 2022-11-30
Provide support for Python 3.10 and 3.11

onegreyonewhite:
Add classifiers about python latest versions + add it to tests.
See also [About distutils](https://peps.python.org/pep-0632/#migration-advice)


onegreyonewhite:
Flower is working under Python 3.10 (tested on my projects).

linpan:
support for Python 3.10

PR #1273, opened 2023-04-06
Support authenticating using Redis usernames

espadrine:
Redis added the AUTH command with only a password at first, and that is what the current Flower code assumes.
Redis added usernames in the AUTH command later, and added support for them [in the redis-py library](https://github.com/redis/redis-py/commit/8df8cd54d135380ad8b3b8807a67a3e6915b0b49#diff-3161b005adfa1f5266a9736bffb5ede9d14e2ab40937cb241bc16ca5647df79dR677).

In this patch, we add support for usernames in Flower.

Note that while the latest release of Kombu [allows many versions of the redis-py package](https://github.com/celery/kombu/blob/v5.2.4/requirements/extras/redis.txt), to support usernames, you need a version that includes the commit linked above. We tested redis-py v4.5.4.


mher:
This change is not backwards compatible

espadrine:
Thanks for discussing it!

What backwards incompatibility does it trigger?

espadrine:
I added tests, in case what you were hinting at was that we want to detect backwards incompatibility on it.

I also tested the following:

1. Launched `docker compose up`;
2. Ran `docker compose run flower celery -A tasks call tasks.add --args '[1, 1]'`;
3. Checked on http://localhost:5555/ that the task was successful.

The default docker-compose.yml configuration does not use a Redis username, which makes me believe there is no backwards incompatibility, but of course there could be something I overlooked.

(As an aside, `tests/call-tasks.sh` has commands with outdated positional parameters which are rejected by celery on the project Dockerfile. I can change that if you wish.)

PR #1277, opened 2023-04-26
Added search arg in List tasks API

snyaggarwal:
Adds all search features like:
- `search="state:FAILED"` -- returns tasks in FAILED state
- `search="args:foo"` -- returns tasks with any arg=foo
- `search=foobar` -- matches with anything foo
and more...

### DIFF
--- a/flower/api/tasks.py
+++ b/flower/api/tasks.py
@@ -524,6 +524,7 @@ List tasks
         received_start = self.get_argument('received_start', None)
         received_end = self.get_argument('received_end', None)
         sort_by = self.get_argument('sort_by', None)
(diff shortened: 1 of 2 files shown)


snyaggarwal:
> Hi @snyaggarwal , I also want to implement the same and trying to find solution for searching arg in List task, found your PR. Do you have any idea when the owner is goona merge it to master branch ?

Not sure!

snyaggarwal:
@mher Can you please approve the workflows to run?

YagoMaia:
If I want to consult more than one state, how would I do it?

For example, I would like to consult all tasks those that are in the Pending, Started, or Received state.

PR #1282, opened 2023-05-13
Remove the favicon and Apple touch icons

pabs3:
The touch icons are from Bootstrap and are no longer used and the favicon is
from Celery but Flower should use its own branding instead of other branding.


mher:
I don't see any issue with using the Celery favicon

PR #1304, opened 2023-07-11
Censor all information from worker options tab

themanifold:
Flower will be default expose all objects in a celery worker in its config tab. See for example this worker:

```
import os
import time
from datetime import datetime

from celery import Celery

app = Celery("tasks",
             broker=os.environ.get('CELERY_BROKER_URL', 'pyamqp://0.0.0.0:5672'),
             )
app.conf.accept_content = ['pickle', 'json', 'msgpack', 'yaml']
app.conf.worker_send_task_events = True
app.conf.config = {"aaa":"bbbb"}

@app.task
def add(x, y):
    return x + y

if __name__ == "__main__":
    app.start()
```

image

By default, there is some censoring of information so that the password for the amqp broker is replaced with stars, though this is no perfect. In practice, you can "smuggle" sensitive information out by using non standard key names (also, I don't know what is doing this fuzzy censoring, so if anyone could point that out, it would be appreciated).

This pull request adds complete censoring of all values in the config tab, but will still display the keys (objects).

Running flower with the `--censor-config` flag will produce the following config tab:
image

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -540,3 +540,10 @@ Default: None
 Sets the URI to which an OAuth 2.0 server redirects the user after successful authentication and authorization.
 
 `oauth2_redirect_uri` option should be used with :ref:`auth`, :ref:`auth_provider`, :ref:`oauth2_key` and :ref:`oauth2_secret` options.
+
+censor_config
+~~~~~~~~~~~~~
+
+Default: False
+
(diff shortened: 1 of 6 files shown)


themanifold:
Is anyone available to review this?

themanifold:
Just bumping this

themanifold:
Bumping this again!

PR #1188, opened 2022-02-15
API Change: Feature - Delete Tasks

0mars:
Example: Delete tasks older than 48hours

```python
import celery
import os
import timedelta
import datetime

broker = os.environ.get('CELERY_BROKER_URL')
app = celery.Celery('tasks', broker=broker)
flower = Flower(capp=app, options=flower_options)

time_delta = timedelta(hours=48)
now = datetime.datetime.now()
delete_before_time = now-time_delta

flower.events.delete_tasks_by_time(delete_before_time.timestamp())
```

PR #1310, opened 2023-08-01
Feature - Drop old tasks with max_count or timedelta

denisolyanyuk:
It would be useful to have the ability to drop tasks in flower: in my case there is a task that runs pretty frequent (for example 2 every mins) and it's important to see last 10 runs and would be better to drop all not important tasks.

This solution is similar to this PR https://github.com/mher/flower/pull/1188 but I think my approach is better.
You need to add one list in  flowerconfig.py and that's all: 

`limit_tasks_by_type=[
    {'type': 'test_task', "max_count": 10, "timedelta": timedelta(minutes=10)},
]`

It will keep 10 newest test_task and drop everything else and drop any test_task older that 10 minutes

### DIFF
--- a/flower/app.py
+++ b/flower/app.py
@@ -62,7 +62,9 @@ class Flower(tornado.web.Application):
             enable_events=self.options.enable_events,
             io_loop=self.io_loop,
             max_workers_in_memory=self.options.max_workers,
-            max_tasks_in_memory=self.options.max_tasks)
+            max_tasks_in_memory=self.options.max_tasks,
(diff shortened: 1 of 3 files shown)


max-wittig (approved):

max-wittig:
@mher Could you take a look at this one? That seems to be a missing feature in Flower.

premchalmeti:
This is required. @mher any plans on merging this?

Eger37 (approved):

PR #1324, opened 2023-09-13
fix pool size value is undefined

daydaychen:
`$('#pool-size option:selected').html()` is undefined，grow and shrink action alway is 1


ap-leemurus:
It's too simple fix, but stupid bug. Let's merge pls.

daydaychen:
Hi @mher! When will this change be released?

PR #1327, opened 2023-09-25
add ssl support for redis with sentinel

cyberjunk:
flower can already connect to sentinel using tls by setting `sentinel_kwargs` in `broker_options` accordingly, e.g.:
```
'sentinel_kwargs': { 'ssl': True, 'ssl_cert_reqs': ssl.CERT_NONE }
```

to enable tls to sentinel without sending any client certificate or validating the server certificate.
this will be passed to the Sentinel() constructor and works fine for the connection from flower to sentinel.

however, this is only one of the two connections made from flower in a sentinel+redis setup...
the next connection which is made to the returned redis master was always without tls.

to make the connection to redis also use tls, one has to pass the according ssl related [connection_kwargs](https://github.com/redis/redis-py/blob/cc4bc1a544d1030aec1696baef2861064fa8dd1c/redis/sentinel.py#L215) to Sentinel() constructor.

I adapted the code from the existing encrypted redis-only case by reusing `broker_use_ssl` settings in the sentinel+redis setup to configure whether the connection to redis should be made using tls or not. checking if `broker_use_ssl` is defined, and if so, set `ssl` to `True` and inject the parameters provided in `broker_use_ssl` for the connection from flower to redis. this way flower can be configured to use tls on both, the sentinel AND the redis connection using existing configuration settings and it's working fine for me now...

PS: This also fixes a `500` error for redis+sentinel on `/broker` route with a server error `104 - Connection reset by peer` indicating that connection to redis was made without tls.

### DIFF
--- a/flower/utils/broker.py
+++ b/flower/utils/broker.py
@@ -159,7 +159,7 @@ class RedisSentinel(RedisBase):
     def __init__(self, broker_url, *args, **kwargs):
         super().__init__(broker_url, *args, **kwargs)
         broker_options = kwargs.get('broker_options', {})
-        broker_use_ssl = kwargs.get('broker_use_ssl', {})
+        broker_use_ssl = kwargs.get('broker_use_ssl', None)
         self.host = self.host or 'localhost'
         self.port = self.port or 26379
         self.vhost = self._prepare_virtual_host(self.vhost)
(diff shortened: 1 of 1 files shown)


mher:
Thanks for the pull request! Can you add unit tests?

PR #1333, opened 2023-10-14
Enable built-in state saving for Tasks datatable

### LINKED ISSUE #1215: More consistent navigation by keeping pagination settings alive between page visits
**Is your feature request related to a problem? Please describe.**
Tasks navigation is inconsistent; *page number*, *number of entries per page*, *selected sorting options* and *search terms* get reset after opening a task and then navigating back (using the browser's back button, because flower does not provide a link back to the index).

**Describe the solution you'd like**
It would help if the browser's location href is updated whenever any of the pagination/filtering/ordering options change. This not only provides more sensible navigation but also allows users to share links to tasks pages with specific pagination settings.

fretman92:
Hello @mher and thanks for this project

Resolve #1215

Relies on https://datatables.net/examples/basic_init/state_save.html

This allows to navigate to a specific task, then navigate back (press browser back) -> it restores the main tasks table state where it was (current page, current sort, ...) instead of going back to first page / initial state.

Maybe the other datatables could benefit as well, but the main use of this app is mainly to browse the list of task of investigate some one by one without having to find where we were each time

Very limited side effect (no external plugin, built-in option) and it keeps the url clean. (sadly no url bookmarking but not very useful for a always evolving list of tasks)

I considered the deeplinking plugin and stateRestore extension, both would have supported url params but it seemed simpler to just use this native configuration.

Thanks for the review (should be quick hopefully ;)


fretman92:
maybe we should have configured an explicit stateDuration (rather than the default 2h using localStorage) https://datatables.net/reference/option/stateDuration
even stateDuration:-1 to use sessionStorage (ie related to the current tab lifecyle) would allow users to quickly reset their saved state by simply opening another tab.
Let's test out how it behaves though, if we get some negative feedback, `stateDuration:-1` will probably appease the crowd ^^

PR #1343, opened 2023-12-14
Adding Auth0 as another option for login

smcl:
I've added an option to login using Auth0 via a new `Auth0LoginHandler`. It's ultimately quite similar to Okta (who are coincidentally their parent company), but with slightly different authorization, token and userinfo endpoints. In addition there are a couple of small changes I made (github's documentation link ended in .htm instead of .html, and Okta's section contained a small unprintable character).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,7 +33,7 @@ Features
     - Revoke or terminate tasks
 - Broker monitoring
     - View statistics for all Celery queues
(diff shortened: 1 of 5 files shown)


smcl:
Damn I only *just* looked now for previous PRs relating to this. Seems you aren't interested in having other providers in the repo? Shame, I'll close this.

PR #1351, opened 2024-01-22
Use celery app timezone

### LINKED ISSUE #1132: Flower show wrong time if timezone set
**Describe the bug**
My local time is UTC+8, but I can not got correct local time in flower view.

My docker settings:

```
  flower:
    image: mher/flower
    environment:
      - TZ=Asia/Shanghai
      - DEBUG=1
      - CELERY_BROKER=redis://redis:6379/0
      - CELERY_BACKEND=redis://redis:6379/0
      - CELERY_TIMEZONE=Asia/Shanghai
      - CELERY_ENABLE_UTC=True
```
and the same in worker docker.

task request at local time `21:00` shown `05:00` in flower. it has a 2 * 8 hours different.

at this time, I checked task info in redis and got:
```
In [9]: datetime.utcfromtimestamp(1627824019.4668603)
Out[9]: datetime.datetime(2021, 8, 1, 13, 20, 19, 466860)
```
which is correct, UTC+8 's 21:00 is UTC 's 13:00.

so why when flower show a utc time in UTC+8 not plus 8 but minus？

**To Reproduce**
set both worker and flower with `CELERY_TIMEZONE`, request a task, and got wrong time show in flower.

**Expected behavior**
right time.

**System information**

```
flower   -> flower:1.0.0 tornado:6.1 humanize:3.10.0
software -> celery:5.1.2 (sun-harmonics) kombu:5.1.0 py:3.9.6
            billiard:3.6.4.0 py-amqp:5.0.6
platform -> system:Linux arch:64bit
            kernel version:5.4.0-80-generic imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:disabled

deprecated_settings: None
```

### LINKED ISSUE #1163: Substantial drift from celery worker
**Is your feature request related to a problem? Please describe.**
Substantial drift from celery worker when celery is running from different container where django has setup for own timezone
Tried with CELERY_TIMEZONE and it didn`t resolve the issue, difference in times still exists.

**Describe the solution you'd like**
Add installation of `tzdata` which gives a possibility set correct timezone in container by setting `TZ` environment.
Fix is provided in #1162

woutdenolf:
Closes #1132

## Changes

- Introduce the flower config parameter `browser_local_time` (bool) to show the time in the Celery app timezone or the browser local timezone.
- When using the Celery app timezone, respect the Celery option `enable_utc` (when Celery option `timezone` is missing the celery app uses UTC or the system TZ).
- The timezone is shown as a tooltip when hovering over the time.
- The full time + timezone is shown as a tooltip when hovering over the time when `natural_time=True`.

## Implementation

The task _started_ and _received_ times in the task table are derived from the raw celery events which provide times as number of seconds passed since the _epoch_. 

https://github.com/mher/flower/blob/58136bf26bb641082d31a376306b73ef5047b80f/flower/views/tasks.py#L117

This code takes `timezone` into account but not `enable_utc`.

https://docs.celeryq.dev/en/stable/userguide/configuration.html#time-and-date-settings

Instead of using the time and date settings from the celery configuration directly, this PR proposes to use the timezone provided by the celery app

https://github.com/celery/celery/blob/8f389997887232500d4aa1a2b0ae0c7320c4c84a/celery/app/base.py#L1360

This allows using an explicit timezone (e.g. `timezone = "Europe/Paris"`) or the local timezone (`enable_utc = False`).

### DIFF
--- a/flower/views/tasks.py
+++ b/flower/views/tasks.py
@@ -102,12 +102,19 @@ class TasksView(BaseHandler):
 
         if not app.options.browser_local_time:
             # Append Celery app timezone in IANA format
+
+            timezone = None
+
             if capp.timezone:
                 if isinstance(capp.timezone, LocalTimezone):
-                    timezone = get_localzone()
+                    try:
+                        timezone = get_localzone()
+                    except Exception as ex:
+                        logger.warning("Failed to retrieve local timezone (%s): %s", type(ex).__name__, ex)
                 else:
                     timezone = capp.timezone
-            else:
+
+            if timezone is None:
                 timezone = ZoneInfo("UTC")
 
             time = f'{time}-{timezone}'
--- a/flower/views/tasks.py
+++ b/flower/views/tasks.py
@@ -1,7 +1,10 @@
 import copy
 import logging
 
+import pytz
 from tornado import web
+from tzlocal import get_localzone
+from celery.utils.time import LocalTimezone
 
 from ..utils.search import QuerySyntaxError
 from ..utils.tasks import as_dict, get_task_by_id, search_tasks
@@ -96,8 +99,18 @@ class TasksView(BaseHandler):
         capp = self.application.capp
 
         time = 'natural-time' if app.options.natural_time else 'time'
-        if capp.conf.timezone:
-            time += '-' + str(capp.conf.timezone)
+
+        if not app.options.browser_local_time:
+            # Append Celery app timezone in IANA format
+            if capp.timezone:
+                if isinstance(capp.timezone, LocalTimezone):
+                    timezone = get_localzone()
+                else:
+                    timezone = capp.timezone
+            else:
+                timezone = pytz.utc
+
+            time = f'{time}-{timezone}'
 
         self.render(
             "tasks.html",
--- a/flower/views/tasks.py
+++ b/flower/views/tasks.py
@@ -1,7 +1,7 @@
 import copy
 import logging
+from zoneinfo import ZoneInfo
 
-import pytz
 from tornado import web
 from tzlocal import get_localzone
 from celery.utils.time import LocalTimezone
@@ -108,7 +108,7 @@ class TasksView(BaseHandler):
                 else:
                     timezone = capp.timezone
             else:
-                timezone = pytz.utc
+                timezone = ZoneInfo("UTC")
 
             time = f'{time}-{timezone}'
 
--- a/tests/unit/views/test_tasks.py
+++ b/tests/unit/views/test_tasks.py
@@ -406,4 +406,4 @@ class TasksTimeZoneTest(AsyncHTTPTestCase):
         r = self.get('/tasks')
         self.assertEqual(200, r.code)
         body = r.body.decode()
-        self.assertIn('time', body)
+        self.assertIn('value="time"', body)
--- a/tests/unit/views/test_tasks.py
+++ b/tests/unit/views/test_tasks.py
@@ -386,10 +386,17 @@ class TasksTimeZoneTest(AsyncHTTPTestCase):
         self._app.capp.conf.enable_utc = False
         self._app.options.browser_local_time = False
 
+        try:
+            expected_tz = get_localzone()
+        except Exception:
+            expected_tz = None
+        if expected_tz is None:
+            expected_tz = 'UTC'
+
         r = self.get('/tasks')
         self.assertEqual(200, r.code)
         body = r.body.decode()
-        self.assertIn(f'time-{get_localzone()}', body)
+        self.assertIn(f'time-{expected_tz}', body)
 
     def test_browser_local_time(self):
         del self._app.capp.timezone  # clear cached property
--- a/tests/unit/views/test_tasks.py
+++ b/tests/unit/views/test_tasks.py
@@ -2,6 +2,7 @@ import json
 import time
 
 from celery.events import Event
+from tzlocal import get_localzone
 
 from flower.events import EventsState
 from tests.unit import AsyncHTTPTestCase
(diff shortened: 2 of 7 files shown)


vabee4489:
58才でさそり座♏ さそり座♏は日本🗾出来る 電気で日本🗾列島ライン通商都市出来るweb3オークション　第九は生存権番号が出せるはweb3級だからかIDよろしくで離れ日本。4489Hiroko　Companyで電波の国は通商都市で決まった
角松さんの生存権??TRUMPのとこだ。
10000万円の岩盤使用料は支払えだな

woutdenolf:
@mher Let me know if you need something. The main idea here is the use `capp.timezone` instead of `capp.conf.timezone` so that we can take `enable_utc = False` into account.

auvipy (changes requested):
is it possible to add some tests to verify this?

woutdenolf:
I added a test. copilot broke the PR so I fixed that as well.

woutdenolf:
I will try to add a test for `enable_utc` as well.

woutdenolf:
The front-end does not actually show the timezone. 

image

Do you think this is required for this PR or can this be another PR?

flower/views/tasks.py
```diff
@@ -113,8 +116,14 @@ def get(self):
         capp = self.application.capp
 
         time = 'natural-time' if app.options.natural_time else 'time'
-        if capp.conf.timezone:
-            time += '-' + str(capp.conf.timezone)
+        if capp.timezone:
+            # Timezone in IANA format
+            if isinstance(capp.timezone, LocalTimezone):
+                timezone = get_localzone()
```
woutdenolf, line 122: FYI: `str(LocalTimezone())` would give something like `<LocalTimezone: UTC+02>` which is not a valid IANA format.

woutdenolf:
Note that I'm using `lzlocal` because just the UTC offset yields more than one timezone

```python
import pytz
from datetime import datetime

local_now = datetime.now()
local_utc_offset = local_now.astimezone().utcoffset()

matching_timezones = []
for tz_name in pytz.all_timezones:
    tz = pytz.timezone(tz_name)
    if tz.utcoffset(local_now) == local_utc_offset:
        matching_timezones.append(tz_name)

print(matching_timezones)
```

flower/views/tasks.py
woutdenolf: Note that I'm using `lzlocal` because just the UTC offset yields more than one timezone

```python
import pytz
from datetime import datetime

local_now = datetime.now()
local_utc_offset = local_now.astimezone().utcoffset()

matching_timezones = []
for tz_name in pytz.all_timezones:
    tz = pytz.timezone(tz_name)
    if tz.utcoffset(local_now) == local_utc_offset:
        matching_timezones.append(tz_name)

print(matching_timezones)
```

auvipy: interesting

auvipy:
************* Module flower.views.tasks
flower/views/tasks.py:3:0: W0611: Unused datetime imported from datetime (unused-import)

auvipy (changes requested):

tests/unit/views/test_tasks.py
```diff
@@ -1,7 +1,10 @@
 import json
 import time
+from datetime import datetime
 
+import pytz
```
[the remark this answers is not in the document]
woutdenolf: Done. And fixed the linting.

woutdenolf:
Thanks for running the pipeline and sorry for not checking all permutations. The tox.ini is broken. This would fix it:

```
[tox]
envlist =
    {py38,py39,py310,py311}-celery52-tornado60,
    {py38,py39,py310,py311,py312}-celery53-tornado60,
    {py38,py39,py310,py311,py312}-celery54-tornado60,
    lint
skip_missing_interpreters = true

[testenv]
deps =
    -r requirements/dev.txt
    celery52: celery==5.2.*
    celery53: celery==5.3.*
    celery54: celery==5.4.*
    tornado60: tornado==6.0.*
commands =
    python -m tests.unit

[testenv:lint]
deps =
    pylint
commands =
    pylint flower --rcfile .pylintrc
```

Do you think it is worth fixing? If yes I'll make an issue and a PR.

loichuder:
Since #1451 is merged, I guess this is now ready for review again? 

We would love to have this in flower :slightly_smiling_face:

auvipy (approved):

mher:
https://github.com/mher/flower/issues/1132#issuecomment-2453270055 shows how to set the local time zone. I don't think we need an additional configuration option.

woutdenolf:
Setting `timezone` in a `celeryconfig.py` works, but this PR covers much more (see the PR description).

These two have been causing us trouble for years:

- **Browser local time.** Opt-in when the user cares about its own timezone, not the timezone celery happens to be configured in.
- **Showing which zone is in use.** Today the table prints `2024-01-01 12:00:00.000` with no zone anywhere, so you can't tell what you're looking at — that's really why #1132 reads as "Flower subtracts hours". The tooltip fixes that, and with `natural_time`, where the cell only says "2 minutes ago", it also gives you the actual timestamp.

The `enable_utc` part is a smaller fix along the way. With `timezone` unset and `enable_utc = False`, Celery means local time. Flower renders the table twice: the initial HTML goes through `current_app.timezone` and shows local time, then the DataTables refresh re-renders the same rows in JS, where the fallback is a hardcoded `'UTC'`. So the times shift as soon as the page updates.

woutdenolf:
> [#1132 (comment)](https://github.com/mher/flower/issues/1132#issuecomment-2453270055) shows how to set the local time zone. I don't think we need an additional configuration option.

You refer to server-side timezone. The additional `browser_local_time` option refers to client-side timezone.

woutdenolf:
Rebased on master to

- resolve conflicts
- drop two commits related to python 3.8 (flower requires >=3.10 by now)

woutdenolf:
An example of what the user sees in the tooltip when hovering over the time

image

**Which timezone is used**

| `browser_local_time` | Celery `timezone` | Celery `enable_utc` | Timezone used |
|---|---|---|---|
| `False` (default) | set | any (ignored) | the configured zone |
| `False` (default) | unset | `True` (default) | `UTC` |
| `False` (default) | unset | `False` | the Flower host's local zone |
| `True` | any (ignored) | any (ignored) | the viewer's browser zone |

**What is shown**

Example: a task received at `11:30:00` UTC, page viewed 5 minutes later, timezone in use `Asia/Shanghai`

| `natural_time` | Cell | Tooltip |
|---|---|---|
| `False` (default) | `2026-09-06 19:30:00.000` | `5 minutes ago (Asia/Shanghai)` |
| `True` | `5 minutes ago` | `2026-09-06 19:30:00.000 (Asia/Shanghai)` |

Same task with the other zone choices, `natural_time = False`:

| Timezone used | Cell | Tooltip |
|---|---|---|
| `UTC` | `2026-09-06 11:30:00.000` | `5 minutes ago (UTC)` |
| `Europe/Brussels` (Flower host) | `2026-09-06 13:30:00.000` | `5 minutes ago (Europe/Brussels)` |
| `America/New_York` (browser) | `2026-09-06 07:30:00.000` | `5 minutes ago (America/New_York)` |

On master the cell text is the same, but there is no tooltip and no zone shown anywhere on the page. Of the last table only the `UTC` row is reachable: the JS table falls back to a hardcoded `UTC`, so neither the Flower host's local zone nor the viewer's browser zone can be selected.

mher:
> > [#1132 (comment)](https://github.com/mher/flower/issues/1132#issuecomment-2453270055) shows how to set the local time zone. I don't think we need an additional configuration option.
> 
> You refer to server-side timezone. The additional `browser_local_time` option refers to client-side timezone.

If it is a client-side time zone, it should not be configured on the server side. Some clients may want to use the browser time zone, others may not.

woutdenolf:
> If it is a client-side time zone, it should not be configured on the server side.

Indeed but Flower still needs the option (--browser-local-time  ) to show the client-side time zone. Otherwise you only see sever-side timezone.

See the last table of https://github.com/mher/flower/pull/1351#issuecomment-5560969636 for an example.

woutdenolf:
So in terms of parameters we now have these:

- `--natural-time`: Flower display option
- `--browser-local-time`: Flower display option (NEW)
- `timezone`: Celery config
- `enable_utc`: Celery config

woutdenolf:
> Some clients may want to use the browser time zone, others may not.

Sure but the same thing can be said for `--natural-time`.

We could make both `--browser-local-time` and `--natural-time` something the user can toggle in their browser (and the flower-slide flag says what the default is). I can make a follow-up PR if you want.

mher:
> Indeed but Flower still needs the option (--browser-local-time ) to show the client-side time zone. Otherwise you only see sever-side timezone.
> 
> See the last table of [#1351 (comment)](https://github.com/mher/flower/pull/1351#issuecomment-5560969636) for an example.

Why are you adding the time zone to relative-time tooltips? Clients already know what time zone they're in, so there's no need to keep reminding them.

woutdenolf:
> Why are you adding the time zone to relative-time tooltips?

Because it isn't necessarily the client's timezone, it can be the Flower timezone or the Celery configured one.

Lets say you see

> 2026-09-06 07:30:00.000

and the tooltip says

> 5 minutes ago

The user would need to do some math to figure out that the timezone of the absolute time shown is (America/New_York). So that's why I'm showing it in the tooltip. Hope it is clear, timezone stuff can be tricky and confusing. We fell over this so many times.

woutdenolf:
@mher Could you provide feedback on why this was closed and what I can do to be able to see time zones (as tooltips or otherwise) in the frontend?

mher:
I think adding another server-side time option unnecessarily complicates the implementation. The display timezone is a per-viewer preference, and the direction I'd like to take is to move all time formatting from the server to the client, possibly deprecating the natural-time option and letting each client decide how to display time.

Issue #1364, opened 2024-03-28
number_of_prefetched_tasks not decrementing


dpdoughe:
**Describe the bug**
The number_of_prefetched_tasks remains greater than zero and grows slowly after service is under load.

**To Reproduce**
I've seen the number_of_prefetched_tasks slowly creep up (by 1 or 2) counts after periods of heavy load (hundreds of thousands of tasks). I don't have a way in hand to reproduce this as I don't know yet what the root cause might be.

What I think may be going on is that the code here 
https://github.com/mher/flower/blob/58136bf26bb641082d31a376306b73ef5047b80f/flower/events.py#L89
will increment the counter if 3 conditions are met.

While the code here 
https://github.com/mher/flower/blob/58136bf26bb641082d31a376306b73ef5047b80f/flower/events.py#L92
is required to decrement the counter.

I wonder if either i) some condition might fail to match the 4 decrementing criteria, or if there is some scenario where the corresponding event would never get fired at all. 

**Expected behavior**
number_of_prefetched_tasks should decrease as well as increase and potentially get back to zero under periods of low to no load.

**Screenshots**
![GrafanaPrefetchedTasksCounter](https://github.com/mher/flower/assets/23531138/905c9bb3-6bdf-4403-b3fd-8c52d352477b)

**System information**
Output of `python -c 'from flower.utils import bugreport; print(bugreport())'` command

flower   -> flower:2.0.1 tornado:6.4 humanize:4.9.0
software -> celery:5.3.6 (emerald-rush) kombu:5.3.4 py:3.8.10
            billiard:4.2.0 py-amqp:5.2.0
platform -> system:Linux arch:64bit
            kernel version:5.15.0-1053-azure imp:CPython
loader   -> celery.loaders.app.AppLoader
settings -> transport:amqp results:disabled

deprecated_settings: None

dpdoughe:
This seems to be some rare scenario in which the number of decrementations doesn't equal the number of incrementations.

dpdoughe:
My current best guess about what is going on here is that this upwards drift can happen when an `expires` is set on a task.  [According to the docs](https://docs.celeryq.dev/en/stable/userguide/calling.html#expiration) , 

> When a worker receives an expired task it will mark the task as [REVOKED](https://docs.celeryq.dev/en/stable/userguide/tasks.html#std-state-REVOKED)

Since the decrementation logic
[...]

PR #1365, opened 2024-04-01
Decrement number_of_prefetched_tasks if task either starts or expires

dpdoughe:
Attempts to provide a solution to https://github.com/mher/flower/issues/1364


mher:
Fixed in https://github.com/mher/flower/issues/1364

PR #1389, opened 2024-08-01
Fix for amqps

gersh:
The broker should work properly if we are using AMQPS.


symbiotech (approved):

mher:

PR #1400, opened 2024-10-11
Add PKCE to Okta auth

rabbit-aaron:
Okta supports PKCE with Client ID / Client Secret authentication.

In OAuth 2.1 (currently draft), auth code flow is recommended to use PKCE.
https://datatracker.ietf.org/doc/draft-ietf-oauth-v2-1/


flower/views/auth.py
```diff
@@ -268,19 +292,31 @@ def _OAUTH_AUTHORIZE_URL(self):
     def _OAUTH_ACCESS_TOKEN_URL(self):
         return f"{self.base_url}/v1/token"
 
+    @property
+    def _oauth_okta_scope(self):
+        return self.application.options.oauth2_okta_scope.split()
```
rabbit-aaron, line 297: OKTA supports adding different claims to the ID token based on the scopes set.
In our case we needed the "groups" claim and that is configured to require a "groups" scope in the request.

mher:
Okta customizations can be implemented in a custom handler class that subclasses OktaLoginHandler https://flower.readthedocs.io/en/latest/auth.html#custom-authentication-providers

PR #1405, opened 2024-10-30
fix timezone

Baghdady92:
fix timezone


jheld:
So this is fixing the value output to the user? as in it doesn't modify the stored data.

PR #1450, opened 2025-08-16
print_banner: print actual port in case the port is dynamic

### LINKED ISSUE #1449: Flower Shows "Visit me at http://0.0.0.0:0" with --port=0
## Description
When starting Flower with Celery using a dynamic port (`--port=0`), Flower logs the URL as:

```bash
celery -A tasks.app flower --port=0
[I 250816 16:10:10 command:168] Visit me at http://0.0.0.0:0
```

The port in the URL shows as `0`, which is confusing. This happens because `--port=0` tells Tornado to let the OS pick a free port, but Flower prints the log **before Tornado binds to the port**. The actual port is chosen at runtime and isn’t automatically reflected in the log.

## Expected Behavior
The log should display the actual port assigned by the OS, e.g.:

```
[I 250816 16:10:10 command:168] Visit me at http://0.0.0.0:52345
```

woutdenolf:
Closes #1449

When starting flower with `--port=0` the port is dynamically assigned but `print_banner` shows `Visit me at http://0.0.0.0:0`.

- Fix by passing the flower app to `print_banner` instead of the celery app.
- For this to work we need to decouple the start of the server and the start of the event loop. We know the port only after the server was started. In addition the event loop might need to be controlled outside of this class. In the unit test framework for example. So the decoupling is useful for that as well.
- For backward compatibility the coupled start method is kept (same for stop)
    ```python
    class Flower(tornado.web.Application):
        ...
        def start(self):  # for backward compatibility
            self.start_server()
            self.update_workers()
            self.serve_forever()
    ```
- To add a unit test for the dynamic port showing up in the banner, the `AsyncHTTPTestCase` base class needs to be in charge of creating the HTTP server. Before it was `tornado.testing.AsyncHTTPTestCase` doing this. Now we derive from `tornado.testing.AsyncTestCase` and let `Flower` handle the HTTP server. This happens in production anyway so now we cover that as well. In addition this makes the usage in other tests much cleaner.

### DIFF
--- a/flower/app.py
+++ b/flower/app.py
@@ -1,5 +1,6 @@
 import sys
 import logging
+from urllib.parse import quote
 
 from concurrent.futures import ThreadPoolExecutor
 
@@ -37,20 +38,20 @@ class Flower(tornado.web.Application):
 
     def __init__(self, options=None, capp=None, events=None,
                  io_loop=None, **kwargs):
+
         handlers = default_handlers
         if options is not None and options.url_prefix:
             handlers = [rewrite_handler(h, options.url_prefix) for h in handlers]
         kwargs.update(handlers=handlers)
+
         super().__init__(**kwargs)
+
         self.options = options or default_options
         self.io_loop = io_loop or ioloop.IOLoop.instance()
         self.ssl_options = kwargs.get('ssl_options', None)
 
         self.capp = capp or celery.Celery()
         self.capp.loader.import_default_modules()
-
-        self.executor = self.pool_executor_cls(max_workers=self.max_workers)
-        self.io_loop.set_default_executor(self.executor)
 
         self.inspector = Inspector(self.io_loop, self.capp, self.options.inspect_timeout / 1000.0)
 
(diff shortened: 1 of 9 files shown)


tests/unit/test_command.py
```diff
@@ -74,38 +74,55 @@ def test_autodiscovery(self):
...
+                    
+                    port = self._app._get_port()
+                    if port != 5555:
+                        break
+                else:
+                    self.fail(f"Port was 5555 after {max_attempts} attempts")
+
+                print_banner(self._app)
+
+                self.assertIn(f'INFO:flower.command:Visit me at http://0.0.0.0:{port}', cm.output)
```
woutdenolf, line 124: This is the unit test for the dynamic port number showing up in the banner.

flower/command.py
```diff
@@ -158,24 +164,18 @@ def is_flower_envvar(name):
...
-
-        logger.info(
-            "Visit me at http%s://%s:%s%s", 's' if ssl else '',
-            options.address or '0.0.0.0', options.port,
-            prefix_str
-        )
+def print_banner(flower_app):
+    if not flower_app.options.unix_socket:
+        url = flower_app.get_url()
+        logger.info("Visit me at %s", url)
```
woutdenolf, line 170: This is the fix. We get the URL from the Flower app. Before we got it from the Celery app before the server started.

flower/app.py
```diff
@@ -63,33 +64,85 @@ def __init__(self, options=None, capp=None, events=None,
...
+        if self._executor is None:
+            logging.debug("Starting executor...")
+            ctx = self.pool_executor_cls(max_workers=self.max_workers)
+            self._executor = ctx.__enter__()  # pylint: disable=unnecessary-dunder-call
+            self.io_loop.set_default_executor(self._executor)
+
+    def _stop_executor(self):
+        if self._executor is not None:
+            logging.debug("Stop executor...")
+            self._executor.__exit__(None, None, None)
```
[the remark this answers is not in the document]
auvipy: AI reviews are not fully correct always. so we can pass some.

mher:

PR #1455, opened 2025-10-02
feature: state filtering + dark theme

paravozz:
dark_theme

- Adds a compact “Filter by state” dropdown next to the Tasks search box, wiring its selections through the existing datatable state-saving so toggles persist across reloads.
- Introduces an accessible theme switcher in the navbar


paravozz:
duplicate of https://github.com/mher/flower/pull/1455

PR #1456, opened 2025-10-06
feature: state filtering + dark theme

paravozz:
- Adds a compact “Filter by state” dropdown next to the Tasks search box, wiring its selections through the existing datatable state-saving so toggles persist across reloads.
- Introduces an accessible theme switcher in the navbar
496445691-186640b4-fc43-4ce7-b916-7ec38db665c3


mher:

PR #1458, opened 2025-10-24
Display of available tasks and max concurrency for better monitoring of tasks

ImKian:
The Flower UI previously lacked visibility into worker concurrency and available task slots, which are crucial for capacity planning and real-time debugging. To address this, new columns were added to display the maximum concurrency and available slots, enabling users to easily monitor each Celery worker’s execution capacity. This enhancement helps users quickly assess system utilization and respond more effectively to task queuing issues.

New Columns

### DIFF
--- a/flower/static/js/flower.js
+++ b/flower/static/js/flower.js
@@ -443,15 +443,14 @@ var flower = (function () {
             order: [
                 [1, "des"]
             ],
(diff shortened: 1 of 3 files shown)


mher:
Per-worker concurrency and free slots are not something most users need on the workers page and the pool size is already shown on the worker detail page.

Issue #1439, opened 2025-07-18
Security Suggestion: Please upgrade Tornado to 6.4.2 to fix CVE-2024-7592


Spute:
Hi, thank you for maintaining this project!

I'm writing to suggest upgrading the dependency on [Tornado](https://github.com/tornadoweb/tornado) to version 6.4.2, which includes a fix for a recently disclosed security vulnerability:

CVE-2024-7592: Previous versions of Tornado had a performance issue in cookie header parsing, which could result in quadratic time complexity. This allowed a potential denial-of-service (DoS) attack by sending specially crafted requests that consume excessive CPU and block the event loop.

See Tornado’s release notes for more details:
🔗 [v6.4.2 Release Note](https://github.com/tornadoweb/tornado/releases/tag/v6.4.2)

"Parsing of the cookie header is now much more efficient. The older algorithm sometimes had quadratic performance which allowed for a denial-of-service attack..."

To improve performance and ensure application security, I strongly recommend upgrading to the latest version.

Thanks again for your great work! 🙏

PR #1461, opened 2025-11-09
Update tornado's lower bound to the earliest secure version

ulgens:
The latest release of the 5.x series was in 2018 and has never received a security update since then. The latest found security issue of the 6.x series is fixed in 6.5.0

Resolves:
- https://github.com/mher/flower/issues/1439

Depends on:
- https://github.com/mher/flower/pull/1463

I'll rebase the branch when #1463 is merged.


mher:

PR #1462, opened 2025-11-09
Modernize the code for Python 3.8

ulgens:
With the help of pyupgrade

Small cleaning before removing Python 3.8 support (EOL)


ulgens:
@Nusnus This will help with 3.8 removal and https://github.com/mher/flower/pull/1461

Nusnus (commented):
I wonder if we should align with python 3.10 already as 3.9 is EOL too.

mher:

PR #1473, opened 2025-12-26
Added support for amqps protocol

dirkmoors:


mher:
Fixed in https://github.com/mher/flower/releases/tag/v2.1.0

PR #1490, opened 2026-03-03
Fix connection leaks in broker and transport access

ShubhAtWork:
## Summary

Several places call `capp.connection()` without using a context manager, creating AMQP/Redis connections that are never closed. Each leaked connection holds open a socket and file descriptor. Under sustained use (e.g. the Broker view refreshing, or `/api/queues/length` being polled by monitoring), this causes gradual file descriptor exhaustion, eventually leading to `Too many open files` errors that crash the process.

## Leaks fixed

| Location | Problem | Fix |
|---|---|---|
| `Flower.transport` property | Called on every request that checks broker type. Opens connection, reads driver type, discards without closing. | Cache the result (transport type never changes at runtime) + context manager for initial lookup |
| `print_banner()` | `app.connection().as_uri()` opens a connection to format the broker URL for the startup log. Never closed. | Context manager |
| `BrokerView.get()` | Two separate `capp.connection()` calls — one for Broker creation, one for display URL. Neither closed. | Single context-managed connection for both |
| `GetQueueLengths.get()` | `capp.connection().as_uri()` opens a connection to get broker URI. Never closed. | Context manager |
| `WorkersView.get()` | `capp.connection().as_uri()` opens a connection on every non-JSON page render. Never closed. | Context manager |

## Test plan

- [x] Verify existing tests pass — no behavioral change, only resource management
- [x] In long-running deployment, confirm no file descriptor growth over time (`lsof -p <pid> | wc -l`)

### DIFF
--- a/flower/api/tasks.py
+++ b/flower/api/tasks.py
@@ -396,7 +396,9 @@ Return length of all active queues
         if app.transport == 'amqp' and app.options.broker_api:
             http_api = app.options.broker_api
 
-        broker = Broker(app.capp.connection().as_uri(include_password=True),
+        with app.capp.connection() as conn:
+            broker_uri = conn.as_uri(include_password=True)
+        broker = Broker(broker_uri,
                         http_api=http_api, broker_options=self.capp.conf.broker_transport_options,
                         broker_use_ssl=self.capp.conf.broker_use_ssl)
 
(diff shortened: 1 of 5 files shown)


mher:
I suggest doing something like this instead

```
  from functools import cached_property

  class Flower(tornado.web.Application):
      @cached_property
      def broker_uri(self):
          with self.capp.connection() as conn:
              return conn.as_uri()

      @cached_property
      def broker_uri_with_password(self):
          with self.capp.connection() as conn:
              return conn.as_uri(include_password=True)

      @cached_property
      def transport(self):
          with self.capp.connection() as conn:
              return getattr(conn.transport, "driver_type", None)
 ```

PR #1492, opened 2026-03-03
Purge offline workers and clean up Prometheus metric labels

ShubhAtWork:
## Summary

When a Celery worker goes offline permanently (e.g. a scaled-down container, a crashed node, a spot instance termination), its state persists indefinitely in three places:

1. **`EventsState.counter`** — per-worker event counter dict (~1 KB per worker)
2. **`Inspector.workers`** — cached inspect results
3. **Prometheus label series** — every `labels(worker_name, ...)` call creates a time series that persists for the lifetime of the process

In long-running deployments with worker churn (autoscaling, rolling deploys, spot instances), these accumulate without bound, causing monotonically growing memory usage, degraded Prometheus scrape performance, and a slow `/workers` page.

## Problem with the existing `--purge_offline_workers` option

The option already exists in `flower/options.py`, and `WorkersView.get()` uses it to **filter workers from the UI**. But it does NOT actually remove them from memory — the counter, inspector, and Prometheus data continue to grow.

## What this PR does

- Add a periodic timer (runs every `purge_offline_workers` seconds, minimum 10s) that scans for workers offline beyond the configured threshold
- Actually purge dead workers from `state.counter`, `inspector.workers`, and all Prometheus metric label series
- Handle **orphaned entries** — workers that appear in counter/inspector but not in `state.workers` (e.g. from incomplete event sequences)
- Add `PrometheusMetrics.remove_worker_metrics()` — iterates all 6 metric families and removes label series matching the worker name
- Add `Inspector.purge_worker()` — removes cached inspect data for a worker

## Test plan

- [x] `pytest tests/unit/test_app.py tests/unit/test_events.py tests/unit/test_inspector.py` — 14 tests pass
- [x] Deploy with `--purge_offline_workers=600`, verify memory stabilizes after workers scale down
- [x] Verify Prometheus `/metrics` endpoint no longer shows labels for purged workers

### DIFF
--- a/flower/app.py
+++ b/flower/app.py
@@ -1,5 +1,6 @@
 import sys
 import logging
+import time
 
 from concurrent.futures import ThreadPoolExecutor
 
@@ -8,6 +9,7 @@ import tornado.web
 
 from tornado import ioloop
 from tornado.httpserver import HTTPServer
+from tornado.ioloop import PeriodicCallback
 from tornado.web import url
 
 from .urls import handlers as default_handlers
@@ -64,6 +66,7 @@ class Flower(tornado.web.Application):
             max_workers_in_memory=self.options.max_workers,
             max_tasks_in_memory=self.options.max_tasks)
         self.started = False
+        self._purge_timer = None
 
     def start(self):
         self.events.start()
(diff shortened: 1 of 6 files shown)


mher:

PR #1522, opened 2026-09-18
feat: add optional Redis backend for persistent state

ashutoshshrimal:
## Summary

Persistent mode currently stores its state in a local `shelve` file, which does not survive restarts in ephemeral or containerized deployments (e.g. Kubernetes pods with no persistent volume). This adds an **opt-in** `--db-url` option that stores the same state in Redis instead, so state can be shared/persisted externally.

It is purely additive — when `--db-url` is unset, behaviour is byte-for-byte identical to today (local `shelve` file).

## Design

- A single `--db-url` option taking a `redis://` or `rediss://` URL, mirroring how Flower/Celery already model connections (`--broker`, `--broker-api`). Connection details, DB number, and TLS all live in the URL.
- The existing `--db` value is reused as the **Redis key** under which the state is stored (default `flower`).
- Only takes effect together with `--persistent`; falls back to the shelve file when `--db-url` is unset.
- `redis` is an **optional** dependency, imported lazily and exposed as an extra: `pip install flower[redis]`.

## Usage

```
celery -A tasks.app flower --persistent=True --db-url="redis://localhost:6379/0" --db="flower"
```

## Details

- `flower/events.py` — `Events` gains a `db_url` param; `load_state`/`save_state` dispatch to shelf or Redis backends. The Redis client is built once and reused (owns a connection pool). State and per-worker counters are pickled together, and `pickle.dumps` runs before the `SET` so a pickling error preserves the previously stored value.
- `flower/options.py` — new `db_url` option (default `None`).
- `flower/app.py` — passes `db_url` through to `Events`.
- `flower/command.py` — validates the URL scheme at startup (rejects non-Redis schemes) and warns if `--db-url` is set without `--persistent`.
- `setup.py` — `extras_require={'redis': [...]}`.
- `docs/config.rst` — documents the option.
- `tests/unit/test_events.py` — Redis persistence tests using an in-memory fake, so no live server is needed in CI.

## Verification

- `python -m tests.unit` passes (added 3 Redis tests; existing shelve/persistence tests unchanged).
- `pylint flower --rcfile .pylintrc` → 10.00/10.
- `python -m flower --version` works with and without the `redis` package installed.


mher:
Closing since writing the whole state to Redis as one pickled blob on every save is not an acceptable approach.

